PR #8, opened 2020-04-30
remove async-iterator merging from smart-subscriptions

hayes:


packages/plugin-smart-subscriptions/src/utils.ts
```diff
@@ -15,3 +16,48 @@ export function stringPath(path: GraphQLResolveInfo['path']): string {
 
   return String(path.key);
 }
+
+export function subscribeOptionsFromIterator<T, Context extends object = object>(
```
hayes, line 20: util for backwards compatibility with async-iterator based subscription models

PR #16, opened 2020-12-06
Add zod plugin

hayes:


packages/plugin-zod/src/errors.ts
```diff
@@ -0,0 +1,11 @@
+export class BadRequestError extends Error {
```
hayes, line 1: not used, should get rid of this

packages/plugin-zod/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default class ZodPlugin<Types extends SchemaTypes> extends BasePlugin<Types> {
+  name: 'GiraphQLZod' = 'GiraphQLZod';
+
+  constructor(builder: GiraphQLSchemaTypes.SchemaBuilder<Types>) {
+    super(builder, 'GiraphQLZod');
+  }
+
+  wrapOutputField(fieldConfig: GiraphQLOutputFieldConfig<Types>) {
+    return new ZodFieldWrapper(fieldConfig);
```
hayes, line 13: should only wrap fields that have a schema, can check fieldConfig.giraphqlOptions for schema and return null if its missing.

PR #29, opened 2021-02-01
add directives plugin

### LINKED ISSUE #27: Document directives
I'm trying to use directives to implement this rate limiting middleware https://github.com/ravangen/graphql-rate-limit but I'm not sure how to do this, the types are telling me that the first argument for `builder.toSchema({})` accepts an array of directives but how would I then attach these directives to specific fields?

hayes:


packages/core/src/build-cache.ts
```diff
@@ -382,6 +382,29 @@ export default class BuildCache<Types extends SchemaTypes> {
         ...config.extensions,
         giraphqlOptions: config.giraphqlOptions,
       },
+      astNode: {
```
hayes, line 385: this is from testing, and needs to be reverted

packages/plugin-directives/src/global-types.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    export interface ExtendDefaultTypes<PartialTypes extends Partial<TypeInfo>> {
+      Directives: undefined extends PartialTypes['Directives']
+        ? {}
+        : PartialTypes['Directives'] & {};
+    }
+
+    export interface SchemaBuilderOptions<Types extends SchemaTypes> {
+      schemaDirectives?: Record<keyof Types['Directives'], typeof SchemaDirectiveVisitor>;
+      schemaTransforms?: ((schema: GraphQLSchema) => GraphQLSchema)[];
```
hayes, line 43: these can't be applied via a plugin, directives that use schemaTransform will need to manual invoke the transforms on the schema after it is returned.  (remove and document)

packages/plugin-directives/src/mock-ast.ts
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  if (type instanceof GraphQLNonNull) {
+    return { kind: 'NonNullType', type: typeNode(type.ofType) as NamedTypeNode | ListTypeNode };
+  }
+
+  return { kind: 'NamedType', name: { kind: 'Name', value: type.name } };
+}
+
+function valueNode(value: unknown): ValueNode {
```
hayes, line 124: Need to document this hacky solution.  May be better to have all values be nulls.

packages/plugin-directives/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "tools",
+    "visitor",
+    "transforms",
+    "plugin"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "graphql-rate-limit-directive": "^1.2.1"
```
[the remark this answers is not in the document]
hayes: yup

hayes:
merging this into a new 2.0 branch since there is technically a breaking change in the types.  I have a few more breaking changes that are landing soon, so going to batch them all together.

PR #47, opened 2021-05-07
docs(README): fix typo

n1ru4l:


hayes:
Thanks!

hayes:
@n1ru4l I was actually just thinking about asking you if you thought a live-query plugin would make sense for GiraphQL.  I know it will work for any schema, but I could image a few places things could be tied in more directly.  I hadn't really seen anything about live-queries until recently, and came up with something sortof similar ([smart subscriptions](https://giraphql.com/plugins/smart-subscriptions).

n1ru4l:
@hayes Hey, since a live query is more execution-related and not schema-related I wouldn't say that it makes that much sense TBH as I don't see any potential benefits.

What in-memory-live-query-store does is creating a copy of the schema with wrapped resolvers for keeping track of the resources a client consumes (see https://dev.to/n1ru4l/collecting-graphql-live-query-resource-identifier-with-graphql-tools-5fm5 for a more in-depth explanation). You then reschedule execution of queries by calling `liveQueryStore.invalidate("Todo:1")` (e.g. in your mutation resolvers).

hayes:
I kinda understand how it works, what I was imagining was more along the lines of defining custom resource IDs on types or fields, or resources to invalidate on mutations.  Basically just convenience stuff that calls into the store  to register or invalidate stuff.

hayes:
API is probably simple enough it's not worth it for now. Not sure how much value it would add.  I'll need to play with it more to get a better idea if anything seems like it could benefit from better integration.  For a lot of the other plugins the big benefits come from taking advantage of the type system which doesn't really apply here

PR #58, opened 2021-05-23
Update objects.md

delaneyj:
Tiny change.  BTW @hayes, I'm finally back after a year on another project and congrats on the amount that's been accomplished with this library in the interim!  Actively trying to make the core gql tech (+mercurius) on an upcoming project.


hayes:
thanks!

PR #61, opened 2021-05-23
More doc fixes

delaneyj:
1. Lists where showing 1./1./1. because of the breaks caused by the code examples.
1. Arbitrary line breaks in paragraphs

### DIFF
--- a/docs/guide/enums.md
+++ b/docs/guide/enums.md
@@ -19,15 +19,15 @@ Enums can be defined a number of different ways:
    }
 
    builder.enumType(Diet, {
-     name: "Diet",
+     name: 'Diet',
    });
    ```
 
 1. Using an array of strings
 
    ```typescript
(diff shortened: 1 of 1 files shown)


hayes:
Thanks! I think some of the line breaks will probably be added back next time the formatter is run, but shouldn't affect the rendered output.

delaneyj:
what formatter are you using?  When I run with VSCode default markdown support I'm getting quite a few changes, such as converting the single quote to double.

hayes:
Situation is a bit messy right now.  Prettier is what I had intentionally set up.  Unfortunately gitbook reformats everything when you edit things through the UI.  I have been meaning to look into replacing it with docusaurus instead.  I know there are still a bunch of little typos and inconsistencies in the docs, definitely not my strength.  Someone recently volunteered to go through them and help clean things up, but I haven't had time to go over everything with them yet.  Hopefully they will be a lot better soon.

PR #65, opened 2021-05-26
Make relay connection args optional explicitly

### LINKED ISSUE #64: Setting default requiredness for arguments breaks relay plugin functionality
The requiredness for the relay plugin args should be set to false. Currently if you set your default requiredness to true you cannot query for a connection without supplying the full set of arguments

NICHTJ3:
I'm like 90% sure this was the correct place to make this change but if not please let me know 😄 

Fixes #64

### DIFF
--- a/packages/plugin-relay/src/field-builder.ts
+++ b/packages/plugin-relay/src/field-builder.ts
@@ -194,10 +194,10 @@ fieldBuilderProto.connection = function connection(
     type: placeholderRef,
     args: {
       ...fieldOptions.args,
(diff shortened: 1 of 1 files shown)


hayes:
Thanks! This looks good. I'm camping and may not be able to release this for a day or 2, but I'll get a new version out as soon as I can

hayes:
published:
 - @giraphql/plugin-dataloader@2.2.2
 - @giraphql/plugin-relay@2.2.6

PR #80, opened 2021-06-04
Update Smart-Subscription Docs

baristikir:
I updated the dependency for the plugin as it was the wrong one in the documentation.


hayes:
thanks!

PR #93, opened 2021-07-02
docs(plugins/dataloader): copy paste mistake

BengtHagemeister:
The validation plugin is not necessary for the dataloader package. This was probably a copy + paste mistake from the validation plugin docs.


hayes:
thanks!

PR #102, opened 2021-07-17
feature: add authField method

### LINKED ISSUE #101: Change context type based on authScopes check
Not sure if this is possible to solve but I am noticing this challenge.

Give this builder:
```ts
type Context = { user?: User };

 Context: Context;
AuthScopes: {
    public: boolean;
    authenticated: boolean;
};
}>({
  plugins: [ScopeAuthPlugin, ValidationPlugin],
  authScopes: async ({ user, req }) => ({
    public: true,
    authenticated: !!user,
})
```

Then query:
```ts
builder.queryField('organizations', (t) =>
  t.field({
    type: [UserObject],

    authScopes: {
      authenticated: true,
    },
    resolve: (_root, _arg, { user }) => {
      console.log(user) // technically, because of auth scopes, the user should not be undefined 
    }
  })
);
```

Notice that we check `authenticated` authScope where we check for the presence of the user in the context. When using the context within a query, expectation (since we checked it on authScopes) is that `user` is not `undefined`.

hayes:
fixes #101 

 - @giraphql/converter@2.4.4-alpha.0
 - @giraphql/core@2.10.2-alpha.0
 - @giraphql/plugin-dataloader@2.5.6-alpha.0
 - @giraphql/plugin-directives@2.3.4-alpha.0
 - @giraphql/plugin-errors@2.2.2-alpha.0
 - @giraphql/plugin-example@2.2.4-alpha.0
 - @giraphql/plugin-mocks@2.2.4-alpha.0
 - @giraphql/plugin-relay@2.9.2-alpha.0
 - @giraphql/plugin-scope-auth@2.6.0-alpha.0
 - @giraphql/plugin-simple-objects@2.4.4-alpha.0
 - @giraphql/plugin-smart-subscriptions@2.3.4-alpha.0
 - @giraphql/plugin-sub-graph@2.3.4-alpha.0
 - @giraphql/plugin-validation@2.6.3-alpha.0

### DIFF
--- a/packages/converter/CHANGELOG.md
+++ b/packages/converter/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+### 2.4.4-alpha.0 - 2021-07-17
+
+**Note:** Version bump only for package @giraphql/converter
+
+
+
+
+
 ### 2.4.3 - 2021-07-10
 
 **Note:** Version bump only for package @giraphql/converter
--- a/packages/converter/package.json
+++ b/packages/converter/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@giraphql/converter",
-  "version": "2.4.3",
+  "version": "2.4.4-alpha.0",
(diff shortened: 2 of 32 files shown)


mmahalwy:
🥳

PR #113, opened 2021-07-28
relay improvements

### LINKED ISSUE #110: Relay mutation builder return type
I had a situation where I wanted to share an input type between a mutation created with the relay mutation helper and one without. The relay mutation builder currently returns void and the input type string was not resolving, irrespective of load order. Would it be possible to return a structure with the generated object refs or somehow resolve the string resolution?

### LINKED ISSUE #112: resolveArrayConnection and null
The `resolveArrayConnection` helper is current resolving `T` to `Foo | null` without complaint. This will result in a runtime error because `node` is not nullable. As arrays can contain holes due to missing relations, I _think_ the best thing to do is to map null to null edges in the result.

This is my current workaround: 

```typescript
        const connection = resolveArrayConnection({ args }, foos);

        if (connection == null) {
          return null
        }

        // work around bug in resolveArrayConnection
        return {
          ...connection,
          edges: connection.edges.map(edge => edge?.node == null ? null : edge) || []
        }
```

hayes:
fixes #112 
fixes #110 

Successfully published:
 - @giraphql/plugin-dataloader@2.6.0-alpha.0
 - @giraphql/plugin-relay@2.10.0-alpha.0

### DIFF
--- a/docs/plugins/relay.md
+++ b/docs/plugins/relay.md
@@ -331,6 +331,16 @@ The `inputOptions` has a couple of non-standard options:
 The `payloadOptions` object also accepts a `name` property for setting the name of the payload
 object.
 
+You can also access refs for the created input and payload objects so you can re-use them in other fields:
+
+```typescript
+// Using aliases when destructuring lets you name your refs rather than using the generic `inputType` and `payloadType`
+const { inputType: DeleteItemInput, payloadType: DeleteItemPayload } = builder.relayMutationField(
+  'deleteItem',
+  ...
+);
+```
+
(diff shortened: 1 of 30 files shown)


jamiehodge:
Thank you for being so fast and responsive to our feedback.

hayes:
No problem! Thanks for opening these issues, makes things better for everyone.

Noticed one more thing I need to fix tomorrow.  Start and end cursors are extracted from edges, and won't work if the first or last edge is null.  Since all the current helpers are index based, I can just compute the cursors manually without the edge.

PR #116, opened 2021-08-02
Is the doc correct ?

MartinPELCAT:
Do we need to create a GiraffeInput from the exemple above or is it a doc mistake ?


hayes:
Yup, just a typo in the docs. Thanks for the fix!

PR #126, opened 2021-08-06
Use Uncapitalize instead of Lowercase for prisma. 

kesne:
I found out that this was causing issues with PascalCase model names in Prisma. We really just need to convert the first letter to lowercase, not the entire model name.


hayes:
thanks!

PR #142, opened 2021-08-18
Remove beta-notice for zod

benbender:
Remove beta-notice for zod@3 and make it the default as it is stable for quite a while now.


hayes:
Thanks!

PR #190, opened 2021-09-26
Empty option objects

smolattack:
Removed paragraph about empty objects as the example above no longer follows this convention


hayes:
Good call.  I'll need to look at this.  I did find a good solution for this, but I probably should add a note about how this works now.   Options objects will automatically become required of they have any non-optional properties contributed by plugins

PR #191, opened 2021-09-27
coolPerm missing

smolattack:
I'm actually not sure about this change but I don't know how best to ask docs questions.


hayes:
Looks like this probably just got missed in a rename/refactor.  Thanks!

PR #192, opened 2021-09-28
feat: with input proposal

Jomik:


hayes:
This looks interesting!  I think there are a few things I would change with how this works:

1. I think `withInput` doesn't really make sense as a name anymore
2. Rather than creating a new object type with methods, it would probably be better to augment input refs to have these methods so any input object can be turned a field.
3. this probably needs to work for mutation, subscription, object and interface fields as well.
4. for the plugin name maybe something like input-as-field or something?

I'll be working on the withInput plugin soon, and can also look into what changes are needed to make refs extensible

Jomik:
1. I agree that the name does not make sense :D
2. Agreed.
3. Agreed.
4. I am not sure - I do not see it as turning the input ref into a field, because it is more of a shorthand 'usage'. Naming is hard.

hayes:
Just merged an initial version of the with-input plugin, going to close this for now, but if you want to pursue this API further let me know!

PR #219, opened 2021-11-17
tests for validation on input

mmahalwy:
hey @hayes! 

We ran into a validation plugin bug where:

1. if `validate` is on the input object, it does not run unless a `validate` is added to a field on the input object
2. validate can take an array of arrays, but only one is run and the last one (which if succeeds, the whole validation seems to succeed?)

Ad you can see, when we add `validate` to a field, it makes the validate on the input field work again, hence why there is an error in the response. Whereas, if the `validate` on the field is missing, no `validate` is run on the input object.

### DIFF
--- a/packages/plugin-validation/tests/example/schema/index.ts
+++ b/packages/plugin-validation/tests/example/schema/index.ts
@@ -243,6 +243,17 @@ const WithValidationInput = builder.inputType('WithValidationInput', {
     [(args) => args.age === 100, { message: 'Incorrect age given' }],
   ],
 });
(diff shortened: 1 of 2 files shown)


hayes:
Looks like a couple of bugs where I was using the wrong variables... easy fix.

Thanks for adding these tests and reporting the issue!

https://github.com/hayes/giraphql/pull/220/files

mmahalwy:
oh nice! glad it was helpful @hayes. Thanks for the quick turn around!

PR #228, opened 2021-11-22
Removed dead links

smolattack:
Or maybe the guide file was renamed? I'm not sure


hayes:
Thanks for pointing this out!  I just fixed some of these in a separate change.  The docs exist, I think gitbook just generated some weird broken links.  I am planning to migrate off of it when I have more time to finish building a custom docs site.

smolattack:
> when I have more time to finish building a custom docs site.

I don't know where you find all that time to build stuff but good on ya. I'd just use docusaurus but then again, I wouldn't know as I don't have any libraries to document

PR #238, opened 2021-12-07
Update dataloader.md

furkandmrblk:
Changed ts to typescript, since the syntax highlighting isn't working on the GiraphQL docs (https://giraphql.com/plugins/dataloader)


hayes:
Weird, I was sure this used to work. thanks!

PR #295, opened 2022-02-05
fix: export ForbiddenError for re-use

PabloSzx:
In order to be able to skip reporting to Sentry any authentication errors (using https://www.envelop.dev/plugins/use-sentry), it's required to have access to the class of ForbiddenError, which currently is not possible because it was not exported from the index nor it's possible to do deep imports because the `exports` of the package don't allow it:

```json
{
  "name": "@pothos/plugin-scope-auth",
  "version": "3.0.0",
  "description": "A Pothos plugin for adding scope based authorization checks to your GraphQL Schema",
  "main": "./lib/index.js",
  "types": "./lib/index.d.ts",
  "module": "./esm/index.js",
  "exports": {
    "import": "./esm/index.js",
    "require": "./lib/index.js"
  },
```

---

Now in a separate but related note, I would also like to have the freedom to customize the whole error instance used for the authorization, not only message, both on a global config and resolver level, if you agree, I can help create a PR for it

thank you @hayes, it's a very nice project it has been a nice experience learning to use it

### DIFF
--- a/.changeset/brave-forks-kick.md
+++ b/.changeset/brave-forks-kick.md
@@ -1,0 +1,5 @@
+---
+'@pothos/plugin-scope-auth': patch
+---
+
+Export ForbiddenError for re-use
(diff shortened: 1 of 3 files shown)


hayes:
Thanks! Definitely open to making auth errors more customizable.

hayes:
@PabloSzx looked at error customization briefly.  Mostly seems pretty straight forward except:  I would want the error formatting logic to have access to the results of any auth checks were.  This is slightly tricky because its not exposed in an easy to use way right now.

This isn't necessarily a blocker, but if you have a proposal that can be extended with this information in the future I'd be more than happy to help make that happen.

PR #311, opened 2022-02-11
Undeclared stuff in federation docs

smolattack:
I can't figure out where `ReviewType` might come from so haven't attempted to fix


hayes:
Whoops! that definitely is a typo. thanks!

hayes:
Fixed the ReviewType thing too

PR #314, opened 2022-02-12
Object Type mocks

Jomik:
The idea is to be able to mock certain returned objects whenever they are seen in the schema.
```ts
const DroidMock = createObjectMock('Droid', (...) => fakeDroid);

const mockedSchema = builder.toSchema({
  typeMocks: [DroidMock],
});
```
It would be really amazing if our mock resolver could get some sensible arguments, like the id of the object being resolved, if using dataloader/relay node. I sadly do not think that is possible, given that we have to run the resolver that we try to wrap, to figure out if it is an id or a full object, and that is probably not what is wanted in most cases.
It would also be nice if we were able to be a bit more clever or dynamic with the length of the lists that are requested, but I have found no good way for that.

### DIFF
--- a/packages/plugin-mocks/src/global-types.ts
+++ b/packages/plugin-mocks/src/global-types.ts
@@ -1,5 +1,5 @@
-import { SchemaTypes } from '@pothos/core';
-import { ResolverMap } from './types';
+import { ObjectParam, Resolver, SchemaTypes, ShapeFromTypeParam } from '@pothos/core';
+import { Mock, ResolverMap } from './types';
 import { MocksPlugin } from '.';
 
 declare global {
@@ -10,6 +10,19 @@ declare global {
 
     export interface BuildSchemaOptions<Types extends SchemaTypes> {
       mocks?: ResolverMap<Types>;
+      typeMocks?: Mock<Types>[];
+    }
+
+    export interface SchemaBuilder<Types extends SchemaTypes> {
+      createObjectMock: <
+        Shape extends NameOrRef extends ObjectParam<Types>
+          ? ShapeFromTypeParam<Types, NameOrRef, false>
+          : object,
+        NameOrRef extends ObjectParam<Types> | string,
+      >(
+        nameOrRef: NameOrRef,
+        resolver: Resolver<unknown, unknown, Types['Context'], Partial<Shape>>,
+      ) => Mock<Types>;
     }
   }
 }
--- a/packages/plugin-mocks/src/index.ts
+++ b/packages/plugin-mocks/src/index.ts
@@ -1,7 +1,14 @@
 import './global-types';
+import './schema-builder';
 import { GraphQLFieldResolver } from 'graphql';
-import SchemaBuilder, { BasePlugin, PothosOutputFieldConfig, SchemaTypes } from '@pothos/core';
-import { ResolverMap } from './types';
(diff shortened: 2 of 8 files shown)


packages/plugin-mocks/src/global-types.ts
```diff
@@ -10,6 +10,19 @@ declare global {
...
+
+    export interface SchemaBuilder<Types extends SchemaTypes> {
+      createObjectMock: <
+        Shape extends NameOrRef extends ObjectParam<Types>
+          ? ShapeFromTypeParam<Types, NameOrRef, false>
+          : object,
+        NameOrRef extends ObjectParam<Types> | string,
+      >(
+        nameOrRef: NameOrRef,
+        resolver: Resolver<unknown, unknown, Types['Context'], Partial<Shape>>,
```
hayes, line 24: what if you have a `resolveList` third argument that defaults to `(...args) => [resolver(...args)]` (just returning a list of 1 item by default, but lets you customize list behavior if you want)

packages/plugin-mocks/src/index.ts
```diff
@@ -11,13 +18,14 @@ export class MocksPlugin<Types extends SchemaTypes> extends BasePlugin<Types> {
     resolver: GraphQLFieldResolver<unknown, Types['Context'], object>,
     fieldConfig: PothosOutputFieldConfig<Types>,
   ): GraphQLFieldResolver<unknown, Types['Context'], object> {
-    const { mocks } = this.options;
+    const { mocks, typeMocks = [] } = this.options;
+    const { parentType: typeName, name: fieldName, type: outputType } = fieldConfig;
 
     if (!mocks) {
```
hayes, line 24: this will prevent mocks from working if you only have typeMocks

packages/plugin-mocks/src/schema-builder.ts
```diff
@@ -0,0 +1,12 @@
+import SchemaBuilder, { SchemaTypes } from '@pothos/core';
+
+const schemaBuilderProto = SchemaBuilder.prototype as PothosSchemaTypes.SchemaBuilder<SchemaTypes>;
+
+schemaBuilderProto.createObjectMock = function createMock(nameOrRef, resolver) {
+  const name = typeof nameOrRef === 'string' ? nameOrRef : (nameOrRef as { name: string }).name;
```
hayes, line 6: this isn't safe.  You can use `this.configStore.getTypeConfig(nameOrRef).name`

packages/plugin-mocks/src/index.ts
```diff
@@ -37,18 +45,48 @@ export class MocksPlugin<Types extends SchemaTypes> extends BasePlugin<Types> {
...
-    if (typeof fieldMock === 'function') {
-      return fieldMock;
+    if (outputType.kind === 'Interface') {
+      const outputName = (outputType.ref as { name: string }).name;
+      const implementers = this.builder.configStore
+        .getImplementers(outputName)
+        .map((implementer) => implementer.name);
+      const mock = typeMocks.find((v) => implementers.includes(v.name));
+
+      return mock?.resolver ?? null;
```
hayes, line 79: This probably needs a way to make it more configurable....  They way this is written, if you mock any node, any node/nodes query will return that type regardless of what id is being queried. Not sure how to solve this though

PR #345, opened 2022-03-18
fix: explicitly make pageInfo non nullable

P4sca1:
`PageInfo` is nullable when `defaultFieldNullability` is `true`, but [the relay spec indicates, that `PageInfo` must not be nullable](https://relay.dev/graphql/connections.htm#sec-Connection-Types.Fields.PageInfo)

### DIFF
--- a/packages/plugin-relay/src/schema-builder.ts
+++ b/packages/plugin-relay/src/schema-builder.ts
@@ -409,6 +409,7 @@ schemaBuilderProto.connectionObject = function connectionObject(
       pageInfo: t.field({
         ...pageInfoFieldOptions,
         type: this.pageInfoRef(),
+        nullable: false,
         resolve: (parent) => parent.pageInfo,
       }),
       edges: t.field({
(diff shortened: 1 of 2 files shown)


hayes:
Thanks for the PR!

Since this has the potential to change the resulting schema for anyone affected by this bug, I don't feel comfortable making it a patch version change.

Since it is a bug, and the change it introduced is non-breaking from a client perspective, I think we can avoid a major, and just do a minor version bump.

Can you update the changeset to be a minor version, and mention how this can be reverted by setting nullable in pageInfoFieldOptions in the builders relay options

packages/plugin-relay/src/schema-builder.ts
```diff
@@ -409,6 +409,7 @@ schemaBuilderProto.connectionObject = function connectionObject(
       pageInfo: t.field({
         ...pageInfoFieldOptions,
         type: this.pageInfoRef(),
+        nullable: false,
```
hayes, line 412: This should probably be before `pageInfoFieldOptions` so it can be overwritten.

hayes: Actually, I can just do this before I release, fixing a couple other things anyway.

PR #352, opened 2022-03-20
Create docs to add custom context based on auth scopes

willianhf:
As mentioned in the discussion #350, the auth-plugin had the `AuthContexts` feature not covered in docs.


hayes:
Thanks for adding this!

PR #371, opened 2022-04-13
Improve non-strict user experience

### LINKED ISSUE #319: Builder should have type error when not in strict mode
There is likely an easy way w can detect if we are in strict mode and cause the builder to have a type error with a message indicating that strict mode should be enabled.

hayes:
fixes #319 

New experience when strict mode is disabled:

https://user-images.githubusercontent.com/816527/163256976-5af250c8-ce38-4af9-81ec-d7db590c99ef.mov

Several plugins have been updated to work better with strict mode disabled, but user should still have it enabled if possible.

### DIFF
--- a/.changeset/chilled-timers-smile.md
+++ b/.changeset/chilled-timers-smile.md
@@ -1,0 +1,12 @@
+---
+'@pothos/core': minor
+'@pothos/plugin-directives': minor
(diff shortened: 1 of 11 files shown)


notrab:
Nice!

PR #400, opened 2022-05-12
Small fixes in examples / simple-interfaces

andreasthoelke:
This example was very helpful to me, thank you.

I noticed some small inconsistencies.

### DIFF
--- a/examples/simple-interfaces/src/schema.ts
+++ b/examples/simple-interfaces/src/schema.ts
@@ -52,7 +52,7 @@ Comment.implement({
     }),
     post: t.field({
       type: Post,
-      resolve: (comment) => [...Posts.values()].find((post) => post.id === comment.postId),
+      resolve: (comment) => [...Posts.values()].find((post) => post.id === comment.postId)!,
     }),
   }),
 });
--- a/examples/simple-interfaces/src/schema.ts
+++ b/examples/simple-interfaces/src/schema.ts
@@ -76,7 +76,7 @@ builder.queryType({
         take: t.arg.int(),
         skip: t.arg.int(),
       },
-      resolve: (root, {skip, take}) =>
+      resolve: (root, { skip, take }) =>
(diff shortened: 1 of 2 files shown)


andreasthoelke:
Btw, happy to apply the same to the `simple-classes` example (`examples/simple-classes/src/schema.ts`) if I got this right.

hayes:
> Btw, happy to apply the same to the `simple-classes` example (`examples/simple-classes/src/schema.ts`) if I got this right.

That would be great!

examples/simple-interfaces/src/schema.ts
```diff
@@ -48,11 +48,11 @@ Comment.implement({
...
       type: User,
       nullable: true,
-      resolve: (comment) => Users.get(comment.id),
+      resolve: (comment) => [...Users.values()].find((user) => user.id === comment.authorId),
     }),
     post: t.field({
       type: Post,
-      resolve: (comment) => Posts.get(comment.id)!,
+      resolve: (comment) => [...Posts.values()].find((post) => post.id === comment.postId),
     }),
```
hayes, line 56: I think this needs a non-null assertion, or the field needs to be made nullable, looks like there is a type error when running the typescript on this example

andreasthoelke: Right. It's great that typescript/pothos catches this. I had only tested in a different project before where the typescript error did not show, it seems I'm missing a related tsconfig setting.

I think, in the case of the `Comment` type it makes sense to assume that the database schema assures that a comment can only exist with a reference to a post. Whereas a comment might have no (/an anonymous) author. I propose that we thus assert that the `post` field’s `resolve` value is non-null (using the `!` operator) whereas the `author` field is `nullable: true`.

`pnpm run ci` now succeeds.

I'll still add `examples/simple-classes` to this pr.

hayes:
Thanks so much for these fixes!

PR #452, opened 2022-07-06
feat: add otel example

timsuchanek:
Adds an example for Pothos + Zipkin + Open Telemetry.
This example is very minimal and just there to make it easier to get started.
It's mostly based on the docs.


hayes:
thanks! rebased and merged locally.

Issue #427, opened 2022-06-20
[plugin-federation] schema should contain @key directives on entities


leeroyrose:
Using the example graph created for the [unit tests](https://github.com/hayes/pothos/tree/main/packages/plugin-federation/tests/example)

When a graph is introspected that implements `builder.asEntity`, the `@key` directive is missing from the output - as seen in the unit test snapshots [test snapshots](https://github.com/hayes/pothos/blob/main/packages/plugin-federation/tests/__snapshots__/index.test.ts.snap#L45). The expected output would be similar to [this snapshot](https://github.com/hayes/pothos/blob/main/packages/plugin-federation/tests/__snapshots__/index.test.ts.snap#L10)

Though the graph technically resolves it's entities correctly (due to the _service { sdl } query rendering the @key directive), it's making graph composition using the rover cli fail.

Is this something that would be acceptable to add? If so, I can work on a PR. Thank you!

hayes:
This is an issue with graphql.  As you can see in the snapshots when using the printSubGraphSchema method, the directive is defined.   GraphQLs printSchema method doesn't support printing directives.   The solution is to use another printing method. One option is https://github.com/ardatan/graphql-tools/blob/master/packages/utils/src/print-schema-with-directives.ts

There are some long discussions around this in the graphql repo, but it's currently considered working as intended, which is unfortunate, and an issue that should probably be revised now that there is more active development it.

hayes:
If you want to go down the rabbit hole: https://github.com/graphql/graphql-js/issues/1343 is the current state of this issue.

hayes:
We probably should have better docs around how to set up federation and some of these limitations.  I personally don't have much experience with, or use for apollo federation, other than writing this plugin.  Would be happy to take any feedback or suggestions (or contributions) you might have for the API or docs.

hayes:
On an unrelated note, typescript 4.7 had some nice improvements that would allow the federation plugin API to be a lot cleaner, but would require some breaking changes.  This is part of the reason why it was not released with a > 1.0 version yet.

leeroyrose:
Hey, thank you for the response. That is quite the rabbit hole! given it's a GQL specific implementation, we were able to work around it for our use case without relying on the output directly from an introspection from the pothos sdl.

[...]

PR #472, opened 2022-07-18
Update federation.mdx wrt. printing schema

saevarb:
I've been stumbling around with apollo yelling at me trying to get federation to work, and it eventually lead me to issue #427. Switching to `printSubgraphSchema` fixed the problem, but I did not try your other suggested alternative.


hayes (approved):
thanks

PR #475, opened 2022-07-19
fix linebreaks in final html

miloofcroton:
I see you're limiting line length, probably in your md editor, which isn't a problem for composing regular paragraphs (md compilers usually ignore single linebreaks), but it gets messed up when you have a number list because it creates second paragraphs under the first line of the number (ie "1.", "2.", etc) instead of adding to the first line.


hayes:
thanks!

PR #477, opened 2022-07-21
Fix sub-graph.mdx

knpwrs:
If I'm understanding the remaining docs for this plugin and the types correctly then the initial example on the subgraphs page is incorrect. This PR fixes that.

### DIFF
--- a/website/pages/docs/plugins/sub-graph.mdx
+++ b/website/pages/docs/plugins/sub-graph.mdx
@@ -58,7 +58,7 @@ const combinedSchema = builder.toSchema({ subGraph: ['Internal', 'Public'] });
 
 ### Object and Interface types:
 
-- `defaultSubGraphsForFields`: Default sub-graph for fields of the type to be included in.
+- `defaultForFields`: Default sub-graph for fields of the type to be included in.
 
 ## Options on Fields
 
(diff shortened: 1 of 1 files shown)


website/pages/docs/plugins/sub-graph.mdx
```diff
@@ -58,15 +58,15 @@ const combinedSchema = builder.toSchema({ subGraph: ['Internal', 'Public'] });
 
 ### Object and Interface types:
 
-- `defaultSubGraphsForFields`: Default sub-graph for fields of the type to be included in.
+- `defaultForFields`: Default sub-graph for fields of the type to be included in.
```
hayes, line 61: defaultSybGraphsForFields is the correct option here.

hayes:
Thanks for working on this!  Looks like I forgot some renames when getting this ready for release.  I think some of the options changed here were correct.  The options available can be seen here: https://github.com/hayes/pothos/blob/main/packages/plugin-sub-graph/src/global-types.ts

website/pages/docs/plugins/sub-graph.mdx
knpwrs: I get an error in my editor for that:

image

It seems like it's not defined?

https://github.com/hayes/pothos/blob/ab8479e5891a7c0ef6f14b70162a370d1b0f1595/packages/plugin-sub-graph/src/global-types.ts#L57-L63=

hayes: This is describing the options available on Object and Interface types.  I can't tell from your screenshot where this error is coming from, but it looks (based in the message) like this is for the options based into the builder, not the options for a specific object or interface.

Objects and interfaces have their options extended here: https://github.com/hayes/pothos/blob/main/packages/plugin-sub-graph/src/global-types.ts#L25

knpwrs: Whoops, my bad! I'll change it back.

hayes:
Thanks for fixing this!

PR #479, opened 2022-07-21
docs(prisma): fix spelling of queries

tmilewski:
You have a really amazing project here! It's been a joy playing around with it these past couple of evenings. I hope to contribute more than just a typo change as I dig deeper!

Thanks!


hayes (approved):
thanks!  Docs are definitely not my string suit, so I do really appreciate these types if fixes.

PR #490, opened 2022-08-01
docs: fix Fix bad formatting of app-layout.mdx

ChromeGG:
Something went wrong with this formatting, again. But I'll fix these issues when I find them again 😄


hayes (approved):
thanks for fixing this!

PR #499, opened 2022-08-07
docs: troubleshooting guide for multiple plugin implementations

### LINKED ISSUE #494: Local plugin failing to reload in Next.js
I'm currently migrating an API from Nexus to Pothos and am loving the experience so far. So firstly thank you for your great work and all the TS magic.

**Issue:**
To ease the transition I require some custom plugins which I'd like to locate inside of a subdirectory.
Now the problem I am facing is that whenever I make changes to anything linked to that API route causing it to rebuild OR I load my schema from another place such as a second GraphQL endpoint, an error stating _Recieved multiple implementations for plugin xyz_ is thrown:

grafik

**Example:**
I've created a minimal example for that based on the example code provided in this repo:
https://github.com/feliskio/pothos-nextjs-error-reload
The problem is demonstrated using both Helix and Apollo.

**Solution idea:**
There seems to be a [variable in place](https://github.com/hayes/pothos/blob/731a5701fc717a28c6ca5a97f7b223a4dca16b34/packages/core/src/builder.ts#L79) that could [allow overwriting registered plugins](https://github.com/hayes/pothos/blob/731a5701fc717a28c6ca5a97f7b223a4dca16b34/packages/core/src/builder.ts#L113) but it looks like there is no way to modify that value. If that check could be disabled during development the problem would most likely be solved - although I'm not sure if that would cause other issues internally.

feliskio:
Resolves #494

### DIFF
--- a/website/pages/docs/guide/troubleshooting.mdx
+++ b/website/pages/docs/guide/troubleshooting.mdx
@@ -36,6 +36,14 @@ Ensure that there is only 1 version of each pothos package, and that they are bo
 node_modules directly. Pothos plugins import classes from `@pothos/core` to add plugin specific
 methods to the class prototypes
 
+### Recieved multiple implementations for plugin
+
+By default, Pothos doesn't allow multiple plugin registrations with the same name. During development,
+it can be helpful to disable this check by setting `SchemaBuilder.allowPluginReRegistration = true`.
+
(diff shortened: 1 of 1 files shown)


hayes (approved):
Thanks for adding this!

PR #525, opened 2022-08-12
fix: respect options.complexity.complexityError

P4sca1:
Right now the complexityError can only be set using the builder options.


hayes (approved):
Thanks for the fix!

Issue #534, opened 2022-08-15
New Pothos Plugin for EdgeDB


baristikir:
# Introduction
https://www.edgedb.com/docs

Just found out about EdgeDB and it looks like a new database which also has its own schema and query language, according to my understanding. However it also contains a node.js client for type generations of the underlying database schema, which could be  ideally  defined in a `.esdl` file in projects directory. 

The node.js client actually splits into two different libraries, providing a `Driver API` for executing db queries and a `Query Builder API` to write typesafe queries. For now I would say that the `Query Builder API` would be the entry point, focusing on typesafe integration of schema types rather than providing any performance improvements or query optimizations (also optimizations aren't really documented).

The `QueryBuilder` gets generated from `edgeDB-js` lib and looks mappable to a pothos object. The Pothos plugin API would be actually pretty similar to the prisma plugin, providing advantages of object declartion without explicitly typing the `objectRef`.

But I would say that the biggest benefit could be from the GraphQL query, instead loading manually the schema fields or everything at once, we could determine which were requested from the query and construct the `select` out of it.

```typescript
builder.queryType({
  fields: (t) => ({
    me: t.field({
      type: User, // simple objectRef<User> for db schema
      resolve: async (root, args, ctx, info) => {
            const db_query = await e
                .select(e.User, (user) => {
                    // Manually querying fields
                    name:true,
                    email: true,
                    // Or query everything
                    //  ...e.User["*"],
                    filter: e.op(user.id, '=', ctx.user.id),
                });

            return await db.run(db_query);
      }
    }),
  }),
});

builder.queryType({
  fields: (t) => ({
    // Define a field that issues an edgedb query
    me: t.edgeDBField({
      type: 'User',
      resolve: async (query, root, args, ctx, info) => {
            const db_query = await e
                .select(e.User, (user) => {
                    // automatically resolving queried fields
                    ...query,
                    filter: e.op(user.id, '=', ctx.user.id),
                });

            return await db.run(db_query);
      }
    }),
  }),
});

```

## Example
```typescript
// Create an object type based on a edgedb type
// without providing any custom type information
builder.edgeDBObject('User', {
  fields: (t) => ({
    // expose fields from the database schema
    id: t.exposeID('id'),
    email: t.exposeString('email'),
  }),
});

builder.queryType({
  fields: (t) => ({
    // Define a field that issues an edgedb query
    me: t.edgeDBField({
      type: 'User',
      resolve: async (query, root, args, ctx, info) => {
[...]

PR #539, opened 2022-08-19
wip - Pothos EdgeDB Plugin

baristikir:
# WIP EdgeDB Plugin
Related Issue: https://github.com/hayes/pothos/issues/534

Open for any feedback and reviews!


packages/plugin-edgedb/src/types.ts
```diff
@@ -0,0 +1,389 @@
...
+> = Extract<keyof DefaultExports, KeyType>;
+
+export type EdgeDBModelTypes<Types extends SchemaTypes> =
+  Types['EdgeDBTypes']['default'] extends infer ObjectTypeMap
+    ? ObjectTypeMap extends object
+      ? ObjectTypeMap
+      : never
+    : never;
+
+export type EdgeDBModelShape<
```
baristikir, line 57: EdgeDB's Query Builder contains a default which includes all db schema's with their `properties` and `relations`, or named as `links` in edgeDB, of the schemas and can be used to recreate the `PrismaTypes` structure for type safety.

## Example of EdgeDBTypes
```ts
export interface types {
  "std": {
    ...
  };
  "schema": {
    ...
  };
  "cfg": {
    ...
  };
  "default": {
    "Comment": Comment;
    "Follow": Follow;
    "Media": Media;
    "Post": Post;
    "PostMedia": PostMedia;
    "Profile": Profile;
    "User": User;
  };
  "sys": {
     ....
  };
}

export interface Post extends std.$Object {
  "big_int_id": bigint;
  "content"?: string | null;
  "created_at": Date;
  "published": boolean;
  "title": string;
  "updated_at": Date;
  "author"?: User | null;
}
```

baristikir:
Current work only covers the `edgeDBObject` api. Showcase of how the type system could look like for the edgeDB schema. Type system is constructed out of the EdgeDB Query Builder types, specifically from the default export type. Output shape looks pretty familiar to the PrismaTypes shape

Bildschirmfoto 2022-08-19 um 16 09 54

hayes:
This looks like an awesome start!  Getting the type system working is often one of the hardest parts, and it looks like you are making great progress!

I'll try to find some time today to dive a little deeper.  Initial skim looks like there are lots of pieces that are unused or copied from the prisma plugin that don't quite make sense yet, so I'll just ignore those for now and try to look specifically at types for defining objects and links

packages/plugin-edgedb/CHANGELOG.md
```diff
@@ -0,0 +1,340 @@
+# Change Log
```
hayes, line 1: probably don't need this

packages/plugin-edgedb/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "test": "jest --runInBand"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/hayes/pothos.git"
+  },
+  "author": "Michael Hayes",
+  "license": "ISC",
+  "keywords": [
+    "giraphql",
```
hayes, line 31: probably don't need giraphql in keywords anymore, I think pothos had gotten enough traction that there shouldn't be many people still looking for giraphql

packages/plugin-edgedb/src/edgedb-field-builder.ts
```diff
@@ -0,0 +1,189 @@
...
+  ): PothosSchemaTypes.RootFieldBuilder<Types, Shape, Kind>;
+} = RootFieldBuilder as never;
+
+export class EdgeDBObjectFieldBuilder<
+  Types extends SchemaTypes,
+  Model extends
+    | ({ [ModelKey in keyof Model]: Model[ModelKey] extends infer U ? U : never } & {
+        Fields: string | never;
+        Links: {
+          [Key in Model['Fields']]: { Shape: Model['Links'][Key] };
```
hayes, line 36: curious about this shape referencing itself here, what is that used for?

packages/plugin-edgedb/src/edgedb-field-builder.ts
```diff
@@ -0,0 +1,189 @@
...
+    console.log('[debug] link() -> isList: ', isList);
+
+    return this.field({
+      ...(rest as {}),
+      type: isList ? [ref] : ref,
+      // TODO: Descriptions
+      // description: getFieldDescription(this.model, this.builder, name, description),
+      extensions: {
+        ...extensions,
+        pothosEdgeDBFallback:
```
hayes, line 109: the fallback resolver stuff in pothos is one of my least favorite parts.  having a resolver option that only sometimes gets called is really confusing for users, so if this can be avoided for edgeDB that would be cool.  I haven't seen if you have started working on any entity-reloading logic yet.

packages/plugin-edgedb/src/global-types.ts
```diff
@@ -0,0 +1,108 @@
...
+        Name extends EdgeDBTypeKeys<Types>,
+        Interfaces extends InterfaceParam<Types>[],
+        Model extends EdgeDBModelShape<Types, Name>,
+      >(
+        name: Name,
+        options: EdgeDBObjectTypeOptions<Types, Interfaces, Model>,
+      ) => EdgeDBObjectRef<Types, Model>;
+    }
+
+    export interface EdgeDBObjectFieldBuilder<
```
hayes, line 92: I think I noticed that there is something broken about this in the Prisma plugin.  I tried referencing this global type to create a function that accepts a PrismaFieldBuilder, but it wasn't actually compatible with the PrismaFieldBuilder type.  Not sure if this will have the same issue, but something that might be worth checking on.  Not sure if you need this at all if you are not supporting extending EdgeDBObjectFieldBuilder from other plugins initially.

packages/plugin-edgedb/src/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    | ({ [ModelKey in keyof Model]: Model[ModelKey] extends infer U ? U : never } & {
+        Fields: string | never;
+        Links: {
+          [Key in Model['Fields']]: {
+            Shape: Model['Links'][Key];
+          };
+        };
+      })
+    | never,
+  Shape extends object = Model,
```
hayes, line 38: Shape is generally used as the shape of the `parent` arg

packages/plugin-edgedb/src/types.ts
```diff
@@ -0,0 +1,389 @@
...
+  // castonlyscalar = "castonlyscalar",
+  enum = 'enum',
+  object = 'object',
+  namedtuple = 'namedtuple',
+  tuple = 'tuple',
+  array = 'array',
+  range = 'range',
+}
+export type tupleOf<T> = [T, ...T[]] | [];
+export type cardinalityAssignable<C extends Cardinality> = C extends Cardinality.Empty
```
hayes, line 107: this could probably be modeled in a way thats more readable:

```ts
type CardinalityAssignable<C extends Cardinality> =  {
  Empty: Cardinality.Empty
  ... 
  AtLeastOne: Cardinality.One | Cardinality.AtLeastOne | Cardinality.Many
}[C]

**Comment by hayes** [comment]
Curious about your goals here, and what you are looking for.  Is this something you want to get merged in the not too distant future, or more just sharing/collecting feedback while you iterate?

Long term, is this something you want have live in the main repo and have maintained with the rest of the plugins, or something you want to own in your own repo.  I have been meaning to create some docs pages for linking out to community maintained projects.  It might also be interesting do have a model similar to https://github.com/opentracing-contrib where there is an official place for people to add their own plugins/packages/examples etc in a way that is discoverable.

Open to any ideas and suggestions here, and don't feel too strongly about it.  I would say that the bar for what I would expect in this repo would likely be a lot higher, and include things like making sure that this works well with all the other plugins, thorough documentation, test coverage, and solutions for things like aliasing fields, or defining multiple fields that use the same `link` in the database. 

Regardless of how you want to work on this, and where you would like it to live, I am happy to help out when I have time, and give feedback/reviews or answer any questions you have about how the internals of pothos work.   Looks like you are already making a lot of progress and have figured out a lot of the weird patterns needed to make something like this work.

**Comment by baristikir** [comment]
Agreed, this is right now just not enough for a plugin imo. Long term I would love to see it in the main repo as a plugin. But also looking for more features, supporting other plugins and enabling integrations, like relay nodes and so on. A lot of work needs to be done, not finished at all. 

I am not very familiar with the architecture and type system of Pothos, still learning the patterns and experimenting with the types :)

I was porting a lot of the prisma plugins system into this plugin therefore still unused and unnecessary code / type definitions around here. Will clean that up.

File: packages/plugin-edgedb/src/edgedb-field-builder.ts (hunk 5 above)
Comment (baristikir) [reply packages/plugin-edgedb/src/edgedb-field-builder.ts:109]: Haven't really looked into it, just copy left from the prisma plugin. Will take a look into it, we could avoid it. No entity-reloading logic existing yet.

File: packages/plugin-edgedb/src/edgedb-field-builder.ts (hunk 4 above)
Comment (baristikir) [reply packages/plugin-edgedb/src/edgedb-field-builder.ts:36]: I had parent args, Fields and Links in one type which end up really nested. Will need to rethink that and construct it in a clean way, will probably refer more or less to `PrismaTypes`'s structure. This Shape referencing isnt actually useful there indeed.

https://github.com/hayes/pothos/blob/1f9bec0ed08065585a2c35f9e2afa8bc93cc60a4/packages/plugin-prisma/src/types.ts#L37-L53

Curios is there a difference between `Fields` and `RelationName`? 
Looks like they contain the same values.

Comment (hayes) [reply packages/plugin-edgedb/src/edgedb-field-builder.ts:36]: Nope... That seems to be an oversight.  I'll remove `Fields` from the prisma plugin

**Comment by baristikir** [comment]
Will need some feedback on this topic.

Edgedb queries return type depends on the select shape. Even `...e.User["*"]` won't load the whole db schema. That would be a problem for `edgeDBField` resolve functions return type since `Model["Shape"]` won't be the return type at any time. Instead the return type needs to be composed of optional fields. We might not wanna loose the actual nullability of model properties in `Model["Shape"]`, important for `edgeDBObject` field definitions. So defining another `Shape` type as the optional reflection looks like a solution.

### Example
```ts
// EdgeDB Generated QueryBuilder
import e from '../../client';

builder.queryType({
  fields: (t) => ({
    users: t.edgeDBField({
      type: ['User'],
      nullable: true,
      resolve: async (_query, _parent, _args, ctx) => {
        const users = await e
          .select(e.User, (user) => ({
            id: true,
            email: true,
            name: true,
          }))
          .run(db);

        return users;
        //      ^? { id: string; email:string; name: string | null; }[]
        // This would not match with  Users Model["Shape"], comments | posts ... are missing
      },
    }),
  }),
});

interface EdgeDBModelTypes {
  User: {
    Shape: {
      "comments": Comment[];
      "posts": Post[];
      "email": string;
      "name"?: string | null;
      "followers": Follow[];
      "following": Follow[];
      "media": Media[];
      "profile"?: Profile | null;
    }
  }
}
```

baristikir:
Added `ReturnShape` to `EdgeDBModelTypes` to reflect optional fields from `Shape`.

https://github.com/hayes/pothos/blob/f853b76a8a0bc58ec6fa9b4578bb7aa706515958/packages/plugin-edgedb/src/types.ts#L80-L86

https://github.com/hayes/pothos/blob/f853b76a8a0bc58ec6fa9b4578bb7aa706515958/packages/plugin-edgedb/src/global-types.ts#L102-L104

baristikir:
> Added `ReturnShape` to `EdgeDBModelTypes` to reflect optional fields from `Shape`.
> 
> https://github.com/hayes/pothos/blob/f853b76a8a0bc58ec6fa9b4578bb7aa706515958/packages/plugin-edgedb/src/types.ts#L80-L86

Actually supposed to be a deep partial type, updated version:

https://github.com/baristikir/giraphql/blob/8ec9e25d91d12fa9059a8ec0314c8b493944444e/packages/plugin-edgedb/src/types.ts#L72-L88

hayes:
> Edgedb queries return type depends on the select shape. Even ...e.User["*"] won't load the whole db schema. That would be a problem for edgeDBField resolve functions return type since Model["Shape"] won't be the return type at any time. Instead the return type needs to be composed of optional fields. We might not wanna loose the actual nullability of model properties in Model["Shape"], important for edgeDBObject field definitions. So defining another Shape type as the optional reflection looks like a solution.

I think the technically correct approach here would be to model this after the "select" mode from the prisma plugin.  Basically the prisma plugin supports having a default set of selections defined on the "prismaObject", when you do this, the "parent" type for all resolvers us limited to that selection.  You can still "expose" fields that are not part of the selection (which will automatically select them when the expsosed field is queried) .  You can then also define selections on individual fields, which will extend the parent type in the resolver for that field.

You can see an example of this here: https://github.com/hayes/pothos/blob/main/packages/plugin-prisma/tests/example/schema/index.ts#L38-L59

The implementation for this gets a lot more complicated, so not sure if you need to explore this in the initial version.  Took me a few iterations of the prisma plugin before I figured out how to make that work properly.

PR #555, opened 2022-08-22
fix(#553): use the right types when fullTextSearch is enabled

### LINKED ISSUE #553: `fullTextSearch` breaks types
Hello!

Using [`full-text search`](https://www.prisma.io/docs/concepts/components/prisma-client/full-text-search#enabling-full-text-search) like this:

```prisma
generator client {
  provider        = "prisma-client-js"
  previewFeatures = ["fullTextSearch"]
}
```

Breaks generated types:

image

This is because the full-text search feature creates types with different names:

https://github.com/prisma/prisma-engines/blob/6be410e82fb9c9de89d9348ecd3d2d2359cc8aad/query-engine/schema-builder/src/input_types/objects/order_by_objects.rs#L24-L35

I'll submit a pull request offering a workaround in a few minutes

GauBen:
Fixes #553

This is a temporary workaround while full-text search is in preview (since August 2021), and should be removed as soon as it reaches stable (no ETA yet).

`WithAggregation` seems to be disabled by https://github.com/prisma/prisma-engines/blob/1b996e940dc3998d2003ab429f3ac9a3f98a1fbd/query-engine/schema-builder/src/input_types/fields/arguments.rs#L99-L102

### DIFF
--- a/.changeset/nine-readers-rush.md
+++ b/.changeset/nine-readers-rush.md
@@ -1,0 +1,5 @@
+---
+"@pothos/plugin-prisma": patch
+---
+
+Support `fullTextSearch` types (#553)
--- a/packages/plugin-prisma/src/generator.ts
+++ b/packages/plugin-prisma/src/generator.ts
@@ -38,6 +38,11 @@ generatorHandler({
       options.otherGenerators.find((gen) => gen.provider.value === 'prisma-client-js')!.output!
         .value;
 
+    // See https://github.com/hayes/pothos/issues/553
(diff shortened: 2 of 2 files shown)


hayes:
this is awesome, thanks for contributing a fix!

hayes:
Made this check a little safer in #556.  Enabling the preview feature may change behavior across versions, so instead of looking at the features that are enabled, this looks for the various potential names and uses the one that is actually found.  This should allow this to work across old and new versions after this feature leaves preview as well.

Issue #575, opened 2022-08-31
When using Turborepo, @pothos/plugin-prisma generated types stored in wrong location by default


5t111111:
When using Turborepo, @pothos/plugin-prisma generated types will be stored in wrong location by default.

## Detail

When you use @pothos/plugin-prisma in Turborepo monorepo app, auto-generated types will be stored in a wrong location.

Since Tuborepo downloads NPM packages into workspace root's `node_modules`, `@pothos/plugin-prisma` package is also stored in that path, but when you `npx generate` in individual app directory where `schema.prisma` exists, generated types are stored in the app's `node_modules`. This causes an error when you try to import them from the default location at runtime.

<img width="2439" alt="Gamma_and_Gamma" src="https://user-images.githubusercontent.com/333180/187734102-a934e05b-ea70-4a9c-a8f9-889ccbb188d6.png">

Actually it is not such critical problem as there is nice workaround solution. You can configure `output` option of `generator pothos` in `schema.prisma` or you can change import path as well. However, because Prisma client is stored intended location, ideally a plugin-prisma generated types should also be stored in there too.

## Steps to reproduce

1. `npx create-turbo@latest` and select NPM as package manager
2. In boilerplate `apps/web`, Setup GraphQL server along with Prisma, Pothos and @pothos/plugin-prisma 
3. `npx prisma generate` in `apps/web`
3. `npm run dev` and you will find there an error occurs

abakucz:
Similar thing happened to me with (not with Turborepo).

I'd guess that Pothos assumes that `schema.prisma` is located in located in projects projects root and creates `node_modules\@pothos\plugin-prisma\generated.ts` there by default. E.g. my schema is located in `prisma/schema.prisma` ([specified in package.json](https://www.prisma.io/docs/concepts/components/prisma-schema#prisma-schema-file-location)) and the result of `npx prisma generate` is:
```
```

5t111111:
> I'd guess that Pothos assumes that schema.prisma is located in project's root and creates node_modules\@pothos\plugin-prisma\generated.ts there by default.

The default output path of Pothos is `node_modules/@pothos/plugin-prisma/generated.ts`, when running `prisma generate` the base path likely is set to the location where `schema.prisma` lives.

https://github.com/hayes/pothos/blob/6f23f55e8dea15aee3e346e28af6983bbb13b898/packages/plugin-prisma/src/generator.ts#L30

[...]

PR #584, opened 2022-09-02
Fix: add .ts extension to filename for generated prisma fields

5t111111:
Related issue: #575 

Thanks to https://github.com/hayes/pothos/commit/521cde32f3a223cf5479322418a911f4a467f60e , the default location of the file of generated prisma fields is now correctly set, but missing `.ts` extension confuses module resolution.

I think adding `.ts` is also needed to keep consistency as `generated.ts` is expected filename and it is used in several place, e.g. https://github.com/hayes/pothos/blob/7cbbbe7315b2e5cd2275c1ba0c043d6a2d76abb5/packages/plugin-prisma/prisma/schema.prisma#L17

### DIFF
--- a/.changeset/fluffy-grapes-enjoy.md
+++ b/.changeset/fluffy-grapes-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@pothos/plugin-prisma": patch
+---
+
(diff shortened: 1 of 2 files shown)


hayes (approved):
Thanks, good catch!

PR #602, opened 2022-09-17
Add support for inline composite rules to AuthZ plugin

jonnedeprez:
The AuthZ security layer supports a few more ways to configure authorization rules than what's currently available in the AuthZ plugin. 
For example, if you want to allow an admin user access to many endpoints in your API, you would (with the current implementation of the plugin) need to update all AuthZ rules and add an explicit admin role check. But AuthZ also allows you to write [inline composite rules](https://github.com/AstrumU/graphql-authz#inline-composition-rules), with which you could write an 'IsAdmin' rule just once, and use simple boolean logic to add admin access to endpoints without touching the other rule implementations. 
This PR updates the implementation, the tests and the plugin documentation.

### DIFF
--- a/.changeset/curvy-olives-argue.md
+++ b/.changeset/curvy-olives-argue.md
@@ -1,0 +1,5 @@
+---
+'@pothos/plugin-authz': patch
+---
+
+Added support for inline composite rules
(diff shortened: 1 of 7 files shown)


hayes:
Thanks for adding support for this!  Changes look good to me, just need a changeset to be able to release this

jonnedeprez:
Changeset added

PR #625, opened 2022-10-12
rename and update imports from apollo-server and apollo-server-core to @apollo/server

JipSterk:
Apollo released version 4 of their server but onder a new package. 
This pull request will upgrade to the `@apollo/server` package


hayes:
Looks good, thanks!

PR #667, opened 2022-11-13
Make brandWithType return the branded object

XiNiHa:
This lets users to do something like `return brandWithType(await prisma.user.findUnique({ ... }), User)`

### DIFF
--- a/packages/core/src/utils/index.ts
+++ b/packages/core/src/utils/index.ts
@@ -68,15 +68,28 @@ builder.objectType('MyObject', {
   }
 }
 
-export function brandWithType<Types extends SchemaTypes>(val: unknown, type: OutputType<Types>) {
+type BrandWithTypeResult<T, Types extends SchemaTypes> = T extends Record<string, unknown>
+  ? {
+      [K in keyof T & typeof typeBrandKey]: K extends typeof typeBrandKey
+        ? OutputType<Types>
+        : T[K];
+    }
+  : undefined;
+
+export function brandWithType<T, Types extends SchemaTypes>(
+  val: T,
+  type: OutputType<Types>,
+): BrandWithTypeResult<T, Types> {
   if (typeof val !== 'object' || val === null) {
-    return;
+    return undefined as BrandWithTypeResult<T, Types>;
   }
 
   Object.defineProperty(val, typeBrandKey, {
     enumerable: false,
     value: type,
   });
+
+  return val as BrandWithTypeResult<T, Types>;
 }
 
 export function getTypeBrand(val: unknown) {


packages/core/src/utils/index.ts
```diff
@@ -68,15 +68,28 @@ builder.objectType('MyObject', {
   }
 }
 
-export function brandWithType<Types extends SchemaTypes>(val: unknown, type: OutputType<Types>) {
+type BrandWithTypeResult<T, Types extends SchemaTypes> = T extends Record<string, unknown>
+  ? {
+      [K in keyof T & typeof typeBrandKey]: K extends typeof typeBrandKey
```
hayes, line 73: This doesn't look like it would do what you expect.  Joining with an intersection here probably produces never

packages/core/src/utils/index.ts
```diff
@@ -68,15 +68,28 @@ builder.objectType('MyObject', {
...
+    }
+  : undefined;
+
+export function brandWithType<T, Types extends SchemaTypes>(
+  val: T,
+  type: OutputType<Types>,
+): BrandWithTypeResult<T, Types> {
   if (typeof val !== 'object' || val === null) {
-    return;
+    return undefined as BrandWithTypeResult<T, Types>;
```
hayes, line 84: This should probably return val

packages/core/src/utils/index.ts
```diff
@@ -68,15 +68,28 @@ builder.objectType('MyObject', {
...
+      [K in keyof T & typeof typeBrandKey]: K extends typeof typeBrandKey
+        ? OutputType<Types>
+        : T[K];
+    }
+  : undefined;
+
+export function brandWithType<T, Types extends SchemaTypes>(
+  val: T,
+  type: OutputType<Types>,
+): BrandWithTypeResult<T, Types> {
```
hayes, line 82: I'm not sure adding the brand field to the type here makes sense, it's intended as an invisible side effect, I would probably just return T directly

hayes (commented):
Overall I think returning the passed in value could be a decent change, but I probably wouldn't include the brand field as part of the type, because it shouldn't be accessed outside of Pothos.

packages/core/src/utils/index.ts
XiNiHa: Right, just realized that 😂 Maybe it should be an union...

packages/core/src/utils/index.ts
XiNiHa: But in that case the type error still persists (since the field resolver expects the returned value to have the key attached)

hayes: Do you have an example of how you are trying to use this?

XiNiHa: Something like this:
```ts
author: t.field({
  type: UserOrTeam,
  resolve: async (root) => {
    if (root.authorType === 'USER')
      return brandWithType(
        await prisma.user.findUnique({ where: { id: root.authorId } }),
        User
      )
    else
      return brandWithType(
        await prisma.team.findUnique({ where: { id: root.authorId } }),
        Team
      )
  }
})

Comment (hayes) [reply packages/core/src/utils/index.ts:82]: what is the type error?

Comment (XiNiHa) [reply packages/core/src/utils/index.ts:82]: It was about not having property `[typeBrandKey]` in the returned type (I can't attach the actual error message since it's localized)

Comment (hayes) [reply packages/core/src/utils/index.ts:82]: Oh I see, this is somewhat specific to the prisma plugin... and isn't well documented

what you actually want here is to use the existing tooling for this:

```ts
author: t.field({
  type: UserOrTeam,
  resolve: async (root) => {
    if (root.authorType === 'USER')
      return User.addBrand
        await prisma.user.findUnique({ where: { id: root.authorId } })
      )
    else
      return Team.addBrand(
        await prisma.team.findUnique({ where: { id: root.authorId } }),
      )
  }
})
```

hayes: there is an existing `WithBrand` type in the prisma plugin you could copy or move instead.  I think the type used above is still a bit over-complicated

XiNiHa: Ooooh actually `.addBrand` was the one I was looking for. Maybe I should close the PR then :thinking:

saevarb: Hello. After updating Pothos I started running into this error, and I found the solution here. Previously, I had no errors. This is happening when dealing with a union type.

I was under the impression that this sort of thing was handled by `isTypeOf` and friends, and obviously until this update, things worked. What is the reason that this is needed all of a sudden, and why is there not a single mention of this in the documentation here: https://pothos-graphql.dev/docs/plugins/prisma ? The word `brand` doesn't occur once.

hayes: yes, this definitely needs some docs.  Sorry about that.

So with isTypeOf checks, this technically isn't needed.  The issue is that writing good isTypeOf checks for Prisma objects is almost impossible without adding a column to your DB with the type name.

The way the brands work currently is roughly like this:

When you create a union, if the union doesn't have a resolveType resolver, it will check to see if any of the types declared a custom "abstract shape", which is basically a different type to use when the type is returned as part of a union.  Pothos default resolveType checks for a brands on objects that are part of a union as a way to identify them automatically.

This was always how the node/nodes interfaces worked for prisma objects, but the more recent change was intended to make it easier to build your own interfaces/unions that used prisma objects.  Unfortunately I forgot to document them.

If you have working isTypeOf checks, I can see why this change is pretty inconvenient.  The issue is that pothos can optimize selections to only load the fields that were queried, and consistently identifying objects in an isTypeOf check can be tricky because sometimes the only thing selected is the id.

If you have a resolveType check the expected type should not require branding, but for everything else, it should be easy to add, you basically just need to wrap your prisma calls like: `YourType.addBrand(prisma.yourTyoe.findMany(...))`.   Which should also allow you to remove the isTypeOf checks.  If you have a good way of writing isTypeOf checks let me know, this is something I have never found a good way to do

saevarb: Thanks a lot for your detailed reply!

Yes, that's actually a brilliant idea. I've run into the exact same problems you describe with `isTypeOf` checks, and in a couple of places I'm doing some funky stuff to get it working. 

Thanks for all your hard work as well -- pothos is absolutely great, and I really appreciate the thought and engineering that goes into it.

PR #670, opened 2022-11-14
docs: update example for yoga v3

n1ru4l:
We are releasing Yoga v3 soon.

______

- https://github.com/dotansimha/graphql-yoga/issues/1783


hayes:
Thanks for adding these updates!  Looks like there are some type issues to figure out still, but I can also look into those too if they are not obvious.

hayes:
@n1ru4l it looks like importing yoga is causing issues when type-checking the examples.  The examples mostly all have `"skipLibCheck": false,`, and are more sensitive to type issues in dependencies.    The actual error seems to be related to pulling in incompatible types for web workers, but haven't tracked down the exact issue.

Unfortunately we can't just enable `skipLibCheck`, because having that enabled for these examples has helped catch a number of type compatibility issues in the past.

ardatan:
That issue seems to be fixed in 3.0.3!
https://github.com/dotansimha/graphql-yoga/issues/2123#issuecomment-1326292026
Could you run CI @hayes ?

hayes:
Awesome, thank you!

hayes (approved):

Issue #664, opened 2022-11-11
Custom resolver for node/nodes in @pothos/relay


JipSterk:
So its been awhile since [this](https://github.com/hayes/pothos/issues/340#issuecomment-1094509662) issue. But i think it might be possible to provide (some) support of relay and federation in pothos. Since the implementation of the `@sharable` decorator its now possible to resolve fields on multiple services (but im not really sure if it wil stay this way [source](https://github.com/apollographql/federation/issues/1917#issuecomment-1161463975)). None the less i am willing to work up a draft pr where it possible to add a custom `node/nodes` resolver in the `relayOptions` in `@pothos/plugin-relay`.

With custom resolving of (node/nodes) you would be able to decode a global id and resolve to the correct model and load it from the database.
Consider the following 2 schemas:

user-service:
```graphql
directive @inaccessible on ARGUMENT_DEFINITION | ENUM | ENUM_VALUE | FIELD_DEFINITION | INPUT_FIELD_DEFINITION | INPUT_OBJECT | INTERFACE | OBJECT | SCALAR | UNION

directive @link(import: [link__Import], url: String!) on SCHEMA

directive @override(from: String!) on FIELD_DEFINITION

directive @shareable on FIELD_DEFINITION | OBJECT

interface Node {
  id: ID!
}

type Query {
  node(id: ID!): Node @shareable
  nodes(ids: [ID!]!): [Node]! @shareable
}

type User implements Node @key(fields: "id")
{
  id: ID!
  username: String!
  email: String!
}
 
scalar link__Import
```

reviews-service:
```graphql
directive @inaccessible on ARGUMENT_DEFINITION | ENUM | ENUM_VALUE | FIELD_DEFINITION | INPUT_FIELD_DEFINITION | INPUT_OBJECT | INTERFACE | OBJECT | SCALAR | UNION

directive @link(import: [link__Import], url: String!) on SCHEMA

directive @override(from: String!) on FIELD_DEFINITION

directive @shareable on FIELD_DEFINITION | OBJECT

interface Node {
  id: ID!
}

type Query {
  node(id: ID!): Node @shareable
  nodes(ids: [ID!]!): [Node]! @shareable
}

type Review implements Node @key(fields: "id")
{
  id: ID!
  body: String!
  author: User! @provides("username")
}

type User implements Node @key(fields: "id")
{
  id: ID! @external
  reviews: [Review!]
}
  
scalar link__Import
```

the custom resolver could be like the following:
```ts
[...]

PR #675, opened 2022-11-17
add custom resolve option for node/nodes

JipSterk:
A draft for #664 the only thing i do not know is how to implement type inference.
For now i copied one of the tests & examples but i could provide other tests if needed.
Also i added a small bit to the plugins docs but haven't mentioned that a custom resolve for `node/nodes` isn't something most people would want.

### DIFF
--- a/packages/plugin-relay/src/schema-builder.ts
+++ b/packages/plugin-relay/src/schema-builder.ts
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
-import { defaultTypeResolver, GraphQLResolveInfo } from 'graphql';
 import SchemaBuilder, {
   createContextCache,
   FieldRef,
@@ -16,6 +15,7 @@ import SchemaBuilder, {
   SchemaTypes,
   verifyRef,
 } from '@pothos/core';
+import { defaultTypeResolver, GraphQLResolveInfo } from 'graphql';
 import { ConnectionShape, GlobalIDShape, PageInfoShape } from './types';
 import { capitalize, resolveNodes } from './utils';
 
@@ -156,6 +156,7 @@ schemaBuilderProto.nodeInterfaceRef = function nodeInterfaceRef() {
           id: t.arg.id({ required: true }),
         },
         resolve: async (root, args, context, info) =>
+          this.options.relayOptions?.nodeQueryOptions?.resolve?.(String(args.id), context) ??
           (await resolveNodes(this, context, info, [String(args.id)]))[0],
       }) as FieldRef<unknown>,
   );
@@ -172,9 +173,10 @@ schemaBuilderProto.nodeInterfaceRef = function nodeInterfaceRef() {
         ids: t.arg.idList({ required: true }),
       },
       resolve: async (root, args, context, info) =>
-        (await resolveNodes(this, context, info, args.ids as string[])) as Promise<
+        this.options.relayOptions?.nodesQueryOptions?.resolve?.(args.ids as string[], context) ??
+        ((await resolveNodes(this, context, info, args.ids as string[])) as Promise<
           ObjectParam<SchemaTypes>
-        >[],
+        >[]),
     }),
   );
 
--- a/packages/plugin-relay/src/schema-builder.ts
+++ b/packages/plugin-relay/src/schema-builder.ts
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
+import { defaultTypeResolver, GraphQLResolveInfo } from 'graphql';
 import SchemaBuilder, {
   createContextCache,
   FieldRef,
@@ -15,7 +16,6 @@ import SchemaBuilder, {
   SchemaTypes,
   verifyRef,
 } from '@pothos/core';
-import { defaultTypeResolver, GraphQLResolveInfo } from 'graphql';
 import { ConnectionShape, GlobalIDShape, PageInfoShape } from './types';
 import { capitalize, resolveNodes } from './utils';
 
--- a/packages/plugin-relay/src/types.ts
+++ b/packages/plugin-relay/src/types.ts
@@ -1,3 +1,4 @@
+import { GraphQLResolveInfo } from 'graphql';
 import {
   EmptyToOptional,
   FieldKind,
@@ -28,7 +29,6 @@ import {
   ShapeFromListTypeParam,
   ShapeFromTypeParam,
 } from '@pothos/core';
-import { GraphQLResolveInfo } from 'graphql';
 
 export type RelayPluginOptions<Types extends SchemaTypes> = EmptyToOptional<{
   idFieldName?: string;
--- a/packages/plugin-relay/src/types.ts
+++ b/packages/plugin-relay/src/types.ts
@@ -1,4 +1,3 @@
-import { GraphQLResolveInfo } from 'graphql';
 import {
   EmptyToOptional,
   FieldKind,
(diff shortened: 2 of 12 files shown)


packages/plugin-relay/src/schema-builder.ts
```diff
@@ -156,6 +156,7 @@ schemaBuilderProto.nodeInterfaceRef = function nodeInterfaceRef() {
           id: t.arg.id({ required: true }),
         },
         resolve: async (root, args, context, info) =>
+          this.options.relayOptions?.nodeQueryOptions?.resolve?.(String(args.id), context) ??
```
hayes, line 159: this will fallback to the default resolver if the custom resolver returns null or undefined

packages/plugin-relay/src/schema-builder.ts
```diff
@@ -172,9 +173,10 @@ schemaBuilderProto.nodeInterfaceRef = function nodeInterfaceRef() {
         ids: t.arg.idList({ required: true }),
       },
       resolve: async (root, args, context, info) =>
-        (await resolveNodes(this, context, info, args.ids as string[])) as Promise<
+        this.options.relayOptions?.nodesQueryOptions?.resolve?.(args.ids as string[], context) ??
```
hayes, line 176: I think it makes more sense to decode the ids before passing them to the default resolver

hayes: I think maybe the correct thing here would be to define the id(s) args using t.arg.globalID(s), so they are automatically decoded, resolveNodes would need to be updated to expect the ids to be decoded, but that is an easy change.  

The custom resolver should probably just take all the normal resolver args:
If you do that, then I think the better change here would be something like resolve: `this.option.relayOptions?.nodeQueryOptions?.resolve ?? async ...(default resolver here)`.

packages/plugin-relay/src/types.ts
```diff
@@ -49,20 +49,24 @@ export type RelayPluginOptions<Types extends SchemaTypes> = EmptyToOptional<{
...
       [OutputRefShape<GlobalIDShape<Types> | string>],
       FieldNullability<[unknown]>,
       { ids: InputFieldRef<InputShape<Types, 'ID'>[]> },
-      Promise<unknown>[]
+      Promise<Types['Objects']>[]
     >,
     'args' | 'resolve' | 'type'
-  >;
+  > & {
+    resolve?: (ids: string[], context: Types['Context']) => MaybePromise<Types['Objects'][]>;
```
hayes, line 68: you can remove this, and just remove the `resolve` from the `Omit` above if you take all the normal resolve args

packages/plugin-relay/src/types.ts
```diff
@@ -49,20 +49,24 @@ export type RelayPluginOptions<Types extends SchemaTypes> = EmptyToOptional<{
...
+    resolve?: (id: string, context: Types['Context']) => MaybePromise<Types['Objects']>;
+  };
   nodesQueryOptions: Omit<
     PothosSchemaTypes.QueryFieldOptions<
       Types,
       [OutputRefShape<GlobalIDShape<Types> | string>],
       FieldNullability<[unknown]>,
       { ids: InputFieldRef<InputShape<Types, 'ID'>[]> },
-      Promise<unknown>[]
+      Promise<Types['Objects']>[]
```
hayes, line 64: I think `unknown` is the best we can do here.  It's impossible to type check this.  The type you are using there works for object types explicitly defined in builder setup, but won't work for nodes created with any of the `ref` based apis.

packages/plugin-relay/tests/examples/custom-resolve/builder.ts
```diff
@@ -0,0 +1,59 @@
...
+    pageInfoTypeOptions: {
+      description: 'page info type',
+    },
+    clientMutationIdFieldOptions: {
+      description: 'client id output',
+    },
+    clientMutationIdInputOptions: {
+      description: 'client id input',
+    },
+    mutationInputArgOptions: {
```
hayes, line 55: probably don't need to duplicate all these options that aren't being tested by this example

hayes:
awesome, this looks like a great start!  Thanks for working on this.

I think it's actually pretty close.  

> the only thing i do not know is how to implement type inference.

Can you elaborate on what you mean by this?  I am guessing you mean that type-checking the custom resolver?  Unfortunately I don't think this is possible in a way that is actually useful.

> For now i copied one of the tests & examples but i could provide other tests if needed.

I don't think we need a lot more tests, but simplifying the tests so that there aren't so many unused types/options/fields might be good.  Right now the example used for the test has tons of things that make it hard to figure out what is relevant to that specific test


I think one thing that may be missing (or helpful) is a way to fallback to the default resolver.  I think what you really want is a way to handle just specific ids, and then use the default loaders you already defined for most of the resolvers right?  I am not sure what that API would look like, and it might be something to add later, but I imagine it could be quite useful.

packages/plugin-relay/src/types.ts
JipSterk: i could remove `resolve` from `Omit` but then the resolve options is not nullable so users would have to define it. If people upgrade the package it would break their code.

packages/plugin-relay/src/schema-builder.ts
JipSterk: i have tried to implement the custom resolve like this here: 
```typescript
resolve:
  this.options.relayOptions?.nodesQueryOptions?.resolve ??
  (async (root, args, context, info) =>
    (await resolveNodes(this, context, info, args.ids as string[])) as Promise<
      ObjectParam<SchemaTypes>
    >[]),
```
but it would throw an error about the types being incorrect.
Would the following be a good alternative? 
```typescript
resolve: async (root, args, context, info) => {
  if (this.options.relayOptions?.nodesQueryOptions?.resolve) {
    return (await this.options.relayOptions?.nodesQueryOptions?.resolve(
      this,
      args,
      context,
      info,
    )) as Promise<ObjectParam<SchemaTypes>>[];
  }

  return (await resolveNodes(this, context, info, args.ids as string[])) as Promise<
    ObjectParam<SchemaTypes>
  >[];
```

hayes: I'd rather just cast the type of the resolve function than incur additional runtime overhead to fix a type issue.   You can probably just do 

resolve:
```ts
  this.options.relayOptions?.nodesQueryOptions?.resolve as never ??
  (async (root, args, context, info) =>
    (await resolveNodes(this, context, info, args.ids as string[])) as Promise<
      ObjectParam<SchemaTypes>
    >[]),
```

It's a bit of a hack, but getting these types right internally is less important than making sure the public API is properly type-safe and ensuring that the runtime overhead is as low as possible

packages/plugin-relay/src/types.ts
hayes: ah right, you probably still want to match the resolve type to be similar to the original resolve method (all 4 args, plus supporting the full variety of supported return types).  You can probably just look at how resolve is implemented on the original options object

JipSterk:
the return type of resolve is now `unknown` and removed a lot of options on the test. Getting the tests to pass however isn't something i yet succeeded in. i've modified the arg to be of the type `globalID` & `globalIDList` instead of `id` and `idList` but it throws an error [here](https://github.com/hayes/pothos/blob/3dac1abea8a08319af47fd6f3afb3d705d098306/packages/core/src/config-store.ts#L268) because it cannot resolve the ref because its `undefined` any idea what could cause this?

hayes:
@JipSterk you can merge this: https://github.com/JipSterk/pothos/pull/1 to fix the issues with tests not running, just did a small refactor for how resolveNodes works to avoid extra re-encoding of global ids

hayes (approved):
This looks like a good first version!  I'll probably try to add in a way to fall back to the original resolve node (so you can resolve just specific types with a custom resolver, and fallback to the default for everything else).  

I think all thats missing from this PR is a changeset

PR #676, opened 2022-11-18
Update docs to reference `graphql-yoga`

rahrang:
`@graphql-yoga/node` is deprecated ([see here on npm](https://www.npmjs.com/package/@graphql-yoga/node)).
`graphql-yoga` is the suggested replacement. The Pothos example in the [`graphql-yoga` docs](https://the-guild.dev/graphql/yoga-server/docs) also use `graphql-yoga`.

image


hayes:
Thanks for the PR, there is an existing PR for this here: https://github.com/hayes/pothos/pull/670 with some additional changes.  There is currently still a type issue preventing that PR from being merged, but we should have the docs updated soon.

hayes:
docs have been updated!

PR #684, opened 2022-11-25
Use GraphQL Yoga in Deno recipe

ardatan:


hayes:
This is pretty cool!  It looks like deno.land doesn't work with the changeset style tags.  Importing the deno version of Pothos from skypack seems to work though: https://dash.deno.com/playground/loud-dragonfly-84

hayes:
I don't know a lot about the new npm cross comparability stuff they recently added, so not sure if this is still the best way to do this

ardatan:
I think it works. Do you want me to update the example?
https://dash.deno.com/playground/fine-ferret-22

hayes:
This simple cases work decently well, but there are a lot of issues you can run into with dependencies.  Lot's of graphql libraries (including pothos) only work correctly if everything is using the same version of graphql.  Pothos also uses Sybols for a few internal features, which are shared across packages.  The `@pothos/deno` package solves this by creating a single package with all the plugins, but this isn't really ideal either.

I will try to write up something that shows a few safe ways to use pothos in deno.  I think the best option is using esm.sh + import maps, but I want something you can throw in a playground without import maps as well.

hayes:
@ardatan thanks for starting this docs update, it's awesome to see how easy yoga + deno is to use!  I wanted to write a more thorough guide on setting up deno with pothos, and just merged those changes, so I'm going to close this now.

PR #710, opened 2022-12-15
Fix lazy interfaces on nodes

Emilios1995:
When passing the `interfaces` field to `builder.node` as a function that returns an array (rather than the array directly), a runtime error was occurring. 

The reason was that line `223` of the file changed in this PR was using the spread operator on the passed `interfaces`, which crashes if the value is a function. The compiler didn't catch this because the `as` operator was coercing the value as an array.

### DIFF
--- a/.changeset/fresh-eels-shop.md
+++ b/.changeset/fresh-eels-shop.md
@@ -1,0 +1,5 @@
+---
+'@pothos/plugin-relay': patch
+---
+
+Fix a bug where `builder.node` crashed when the `interfaces` was provided as a function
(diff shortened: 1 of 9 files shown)


hayes (approved):
adding a test would be awesome, but not required, thanks for the fix!

Emilios1995:
> adding a test would be awesome, but not required, thanks for the fix!

thanks for the review! 

i just pushed a commit with a small example with a node object with additional interfaces. i then added a test that just snapshots the generated schema. 

is that what you had in mind?

PR #712, opened 2022-12-16
Fix types on connection fields on interfaces

Emilios1995:
**Original Problem:** I noticed this because I was trying to create a connection with additional fields. Something along the lines of:
```typescript
  users: t.connection(
    {
      type: GraphQLUser,
      resolve: (users, args) => {
        const connection = resolveArrayConnection({ args }, users)
        return <typeof connection & { totalCount: number }>{
          totalCount: recipients.length,
          ...connection,
        }
      },
    },
    {
      fields: tc => ({
        totalCount: tc.int({ resolve: x => x.totalCount }),
      }),
    }
  ),

```
This worked fine on object types, but when I tried this on an interface, the resolver for the `totalCount` int field wasn't aware that the return value of the connection's resolve included a `totalCount` field, so I wasn't able to resolve the value for the `int`. 

**Cause:** I realized that the conditional type changed in this diff didn't match interfaces, as fields there have an optional `resolve` property. This is why it was only matching objects, while interfaces were eventually assigned a default connection shape. 

**Solution:** I changed the conditional type to match an optional `resolve?` property which is the common lowest denominator. 

**Note:** I ran `pnpm` build and noticed that the `deno` package was automatically updated. I included the matching `deno` change caused by this PR, but noticed that the `build` command produced two additional changes that were related to other PRs. (One of them being caused by _my_ previous PR.) You might want to run `build` and put all the changes in master to bring `deno` up to date.

### DIFF
--- a/.changeset/twelve-tables-itch.md
+++ b/.changeset/twelve-tables-itch.md
@@ -1,0 +1,6 @@
+---
+'@pothos/plugin-relay': patch
+---
+
+Fix a bug where the `t.connection` helper wasn't correctly inferring the shape of the returned
+connection object when used on interfaces.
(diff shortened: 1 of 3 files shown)


hayes (approved):
Looks good, nice find!

Re: deno
Definitely something that could be improved.  The deno package itself is mostly maintained for backwards compatibility, the new deno docs don't recommend using that package anymore, so keeping it up to date hasn't been critical, but having unrelated changes show up when you run build is not ideal.

hayes:
Just need to add a changeset to get this released

Emilios1995:
> Just need to add a changeset to get this released

Done!

> Re: deno
Definitely something that could be improved. The deno package itself is mostly maintained for backwards compatibility, the new deno docs don't recommend using that package anymore, so keeping it up to date hasn't been critical, but having unrelated changes show up when you run build is not ideal.

Ah, I didn't realize it was deprecated. We could consider adding a pre-commit hook that runs the `deno` command, though it might not be worth it just to keep a deprecated package on sync.

PR #714, opened 2022-12-17
Add Query type ref to type config

tgriesser:
Fixes the following issue:

image

When attempting to refer to `Query` as a ref. This isn't super common, but can be a useful pattern for cases where you want to return `Query` as a type off of a `Mutation` result.

Also fixes the `pnpm coverage` command

There's probably an opportunity to run `builder.queryType` through `builder.objectType` (what [nexus does](https://github.com/graphql-nexus/nexus/blob/1e6b9212d351e4c05504590ad70b65b2ee0c1f7a/src/definitions/queryType.ts#L42-L44)), allowing interfaces and such, since a `Query` object is just a regular object which also happens to be designated as the root type if another isn't specified.

Looking forward to digging in more here to figure out how this can play nicely with Nexus 😄

### DIFF
--- a/.changeset/tidy-needles-hang.md
+++ b/.changeset/tidy-needles-hang.md
@@ -1,0 +1,5 @@
+---
+"@pothos/core": patch
+---
+
+return a Ref from builder.queryType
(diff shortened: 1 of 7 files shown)


hayes (approved):
Lgtm! Thanks

hayes (approved):

PR #759, opened 2023-01-22
fix: remove duplicate word typo

coopbri:
Changed "for for" to "for" in a custom error message.

The typo also appears in `packages/deno/packages/core/config-store.ts`, however it appears this is a generated artifact based on the core package config store file. With that being said, I'm not exactly sure if it needs to be manually fixed in that file as well (I assume not, but let me know).


hayes (approved):
Thanks!

PR #762, opened 2023-01-22
Rewrite the nextjs example to use Yoga

### LINKED ISSUE #761: Change the Next.js example to use Yoga?
Helix seems to be on track for deprecation: https://github.com/dotansimha/graphql-yoga/issues/1351, https://github.com/contra/graphql-helix/issues/251.

I used the Next.js example to start my own project, and only later found out about the Yoga vs Helix situation. Should I make a PR migrating the Next.js example to Yoga?

neongreen:
Fixes #761.

Sorry for the pnpm-lock.yaml noise, I used the same pnpm version as in `package.yaml` but it didn't help. Perhaps you can reset it and push the regenerated lockfile to my branch (assuming that it even matters).

### DIFF
--- a/examples/nextjs/pages/api/graphql.ts
+++ b/examples/nextjs/pages/api/graphql.ts
@@ -1,34 +1,20 @@
-import {
-  getGraphQLParameters,
-  processRequest,
-  renderGraphiQL,
-  sendResult,
-  shouldRenderGraphiQL,
-} from 'graphql-helix';
-import { NextApiHandler } from 'next/types';
+/** GraphQL API entrypoint */
+
+import { createYoga } from 'graphql-yoga';
+import type { NextApiRequest, NextApiResponse } from 'next';
 import { schema } from '../../graphql/schema';
 
-export default (async (req, res) => {
-  const request = {
-    body: req.body as object,
-    headers: req.headers,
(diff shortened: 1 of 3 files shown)


hayes:
Lock file doesn't matter as long as tests pass, Pothos doesn't have any dependencies in published packages, so shouldn't be an issue

examples/nextjs/pages/api/graphql.ts
```diff
@@ -1,34 +1,20 @@
...
+};
 
-    void sendResult(result, res);
-  }
-}) as NextApiHandler;
+export default createYoga<{
+  req: NextApiRequest;
+  res: NextApiResponse;
+}>({
+  schema,
```
hayes, line 18: I know not all existing examples have this, but all examples should have a context factory that at a minimum creates an empty context object.  Some Pothos plugins require a context object, and it's good practice to set this up for any new graphql api

hayes:
Thanks for adding this!

examples/nextjs/pages/api/graphql.ts
neongreen: I agree and I'll add a context factory

hayes (approved):

PR #765, opened 2023-01-24
Make `resolveArrayConnection` return type non-nullable

### LINKED ISSUE #764: `resolveArrayConnection` returns nullable `ConnectionShape`
This had me scratching my head today. :)

I believe this type should be `false`, as `resolveArrayConnection` always returns a connection object:
https://github.com/hayes/pothos/blob/f3df85be74fb5d5b5478178a6e0469313223762f/packages/plugin-relay/src/utils/connections.ts#L140

Am I understanding correctly? I'm happy to open a PR if so.

paulrosania:
Closes #764

### DIFF
--- a/.changeset/fair-wombats-try.md
+++ b/.changeset/fair-wombats-try.md
@@ -1,0 +1,7 @@
+---
+'@pothos/deno': patch
+'@pothos/plugin-relay': patch
+---
+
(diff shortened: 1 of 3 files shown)


hayes (approved):
Thanks for this fix!

paulrosania:
Thanks for Pothos! We are coming from Nexus and pretty excited about it.

PR #768, opened 2023-01-26
Fix typos on website's plugin pages, and in corresponding READMEs

3nk1du:
## Summary
Make website and README changes for plugins:
1. Fix typos in the **SubGraph plugin**, making sure to make the same changes in the corresponding README.
2. Fix typos in the **Mocks plugin**, making sure to make the same changes in the corresponding README.
3. Remove Mocks plugin info from **SubGraph plugin** documentation.

Ignore anything that could change meaning.

### DIFF
--- a/website/pages/docs/plugins/sub-graph.mdx
+++ b/website/pages/docs/plugins/sub-graph.mdx
@@ -77,9 +77,6 @@ const combinedSchema = builder.toSchema({ subGraph: ['Internal', 'Public'] });
   to being part of the same sub-graph as their parent type. Only applies when type does not have
   `defaultForFields` set.
 
-You can mock any field by adding a mock in the options passed to `builder.toSchema` under
-`mocks.{typeName}.{fieldName}`.
-
 ### Usage
 
 ```typescript
--- a/website/pages/docs/plugins/sub-graph.mdx
+++ b/website/pages/docs/plugins/sub-graph.mdx
@@ -14,7 +14,7 @@ export const getStaticProps = () => ({ props: { nav: buildNav() } });
 
 # SubGraph Plugin
 
-A plugin for creating sub-selections of your graph. This Allows you to use the same code/types for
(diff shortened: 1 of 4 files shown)


website/pages/docs/plugins/sub-graph.mdx
```diff
@@ -77,7 +77,7 @@ const combinedSchema = builder.toSchema({ subGraph: ['Internal', 'Public'] });
   to being part of the same sub-graph as their parent type. Only applies when type does not have
   `defaultForFields` set.
 
-You can mock any field by adding a mock in the options passed to `builder.buildSchema` under
+You can mock any field by adding a mock in the options passed to `builder.toSchema` under
```
3nk1du, line 80: I believe you meant to `toSchema` here. Also would you like me to reference the Mocks plugin here, or do you expect this to work out of the box for the SubGraph plugin (I haven't used it yet).

The change would read:

You can mock any field by using the [Mocks plugin](https://pothos-graphql.dev/docs/plugins/mocks) and adding a mock in the options passed to `builder.toSchema` under...

hayes: I think this was accidentally copied from the mocks plugin, and should just be removed.  The sub-graph plugin doesn't have any special relationship with mocking, and probably shouldn't mention it in the docs.

hayes (approved):
Thank you so much for all these fixes, I really appreciate it!  I know these docs were in a pretty rough state

website/pages/docs/plugins/sub-graph.mdx
3nk1du: Good, that makes more sense :)

3nk1du:
> Thank you so much for all these fixes, I really appreciate it! I know these docs were in a pretty rough state

I'm enjoying it. The docs I've read so far have been clear and well written. I don't think I've had to go digging in the code yet for anything, which is surprising. So, compliments to the author!

PR #775, opened 2023-02-02
Add nodeField to relay edgeObject and edgesField to relay connectionObject

TLadd:
Related to https://github.com/hayes/pothos/discussions/774

### DIFF
--- a/packages/plugin-relay/src/global-types.ts
+++ b/packages/plugin-relay/src/global-types.ts
@@ -31,6 +31,7 @@ import {
   GlobalIDInputShape,
   GlobalIDListFieldOptions,
   GlobalIDListInputFieldOptions,
+  GlobalIDShape,
   InputShapeWithClientMutationId,
   NodeFieldOptions,
   NodeListFieldOptions,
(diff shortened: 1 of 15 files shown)


hayes:
This looks great! thanks for adding this.  A test here might be useful.  The easiest way to do this would be to add use the new edgesFeild/nodeField options to add a description to these fields in the existing schemas.  There are existing snapshot tests that veryify the expected schema.  These tests should fail or need to be updated once you add a description, which can verify that these options are being applied correctly

hayes:
A changeset and an update the the docs would also be appreciated, but I can also add those myself later.

packages/plugin-relay/src/global-types.ts
```diff
@@ -355,6 +356,19 @@ declare global {
...
+          Types,
+          {},
+          ObjectRef<{}>,
+          Types['DefaultNodeNullability'],
+          {},
+          GlobalIDShape<Types> | string
+        >,
+        'args' | 'nullable' | 'resolve' | 'type'
+      > & {
+        nullable?: Types['DefaultNodeNullability'];
```
hayes, line 370: I think the nullable for these needs to be removed here since we already have an existing option for that (same for node)

hayes: technically this would have been better, if I had done it this way all along, but since the edgesNullable and nodeNullable options already exist, this will just add confusion

TLadd:
For the docs, I added the fields to the existing example in https://pothos-graphql.dev/docs/plugins/relay#creating-connections. Wasn't sure if that's good or if you'd prefer it to be in its own section at the bottom, or maybe even combined with [Changing nullability of edges and nodes](https://pothos-graphql.dev/docs/plugins/relay#changing-nullability-of-edges-and-nodes).

Also, wasn't sure if publishing the deno package as a minor makes sense or not in the changeset. Let me know if you'd like anything changed from here!

hayes (approved):
Thanks for adding this!

PR #781, opened 2023-02-06
Add option for enforcing type of for globalID inputs

hayes:


packages/plugin-relay/tests/examples/relay/schema/numbers.ts
```diff
@@ -17,11 +17,12 @@ class BatchLoadableNumberThing {
...
 }
 
-builder.node(NumberThing, {
+const NumberThingRef = builder.node(NumberThing, {
   id: {
     resolve: (n) => n.id,
+    parse: (id) => Number.parseInt(id, 10),
   },
-  loadOne: (id) => new NumberThing(Number.parseInt(id, 10)),
+  loadOne: (id) => new NumberThing(id),
```
tgriesser, line 25: 🎉

packages/plugin-relay/src/utils/internal.ts
```diff
@@ -18,10 +19,28 @@ export function internalDecodeGlobalID<Types extends SchemaTypes>(
...
+  types?: string[] | null,
 ) {
-  if (builder.options.relayOptions.decodeGlobalID) {
-    return builder.options.relayOptions.decodeGlobalID(globalID, ctx);
+  const decoded = builder.options.relayOptions.decodeGlobalID
+    ? builder.options.relayOptions.decodeGlobalID(globalID, ctx)
+    : decodeGlobalID(globalID);
+
+  if (types && !types.includes(decoded.typename)) {
+    throw new Error(`ID: ${globalID} is not of type: ${types.join(', ')}`);
```
tgriesser, line 30: A bit off topic from the rest of the work here (which looks great), but something I was thinking about regarding how this might be used as an API validation layer.

Wondering whether it'd be helpful to throw framework logic errors like this with a `PothosError` class which extends `Error`, so it'd be possible to distinguish it from other types of internal application errors that might happen in a resolver which you might not want to expose publicly.

hayes: Yes! this has been on my todo list for a while.  I really want to create a plugable/configurable error system in Pothos so customizing error messages is easy, but having all pothos error sub-class a core PothosError class would be a good and easy to implement start.

hayes:
@tgriesser one complication I just noticed:

Using a combination of `parse` and a globalID arg using `for` works as expected, but for `globalID` fields that do not use `for`, the ID will still be parsed, but we can't infer the ID type.  

I'll have to look at this more closely, but I think the correct thing to do here would be to only parse the ID when the `for` option is set, and just leave it as a string otherwise.

hayes:
Updated the parsing logic to only parse IDs when using `t.arg.globalID` with a `for` option (and when decoding internally before calling methods like `loadOne`).  This ensures that the types match the runtime behavior, and there isn't a good way to get the types to work correctly if you don't specify the expected types.

tgriesser (approved):
Nice, looks great! 

Don't have a strong opinion about it, but just a suggestion that maybe the type change on the errors could be split out/landed in a separate PR so it's clearer what the actual change is here.

PR #787, opened 2023-02-08
docs: add context to scope-auth plugin ordering relative to relay

### LINKED ISSUE #708: Order of plugins: PothosRelayPlugin + ScopeAuthPlugin, changes globalId type
The docs say the ScopeAuthPlugin should be first, but it seems like having auth first creates type issues when using GlobalIds.

Example:

```ts
import builder from './builder';

builder.relayMutationField(
  'test',
  {
    inputFields: t => ({
      orgId: t.globalID({ required: true }),
    }),
  },
  {
    authScopes: (root, args) => {
      console.log('authScopes', args.input.orgId);

      args.input.orgId.typename; // typescript validates

      return true;
    },
    async resolve(root, args, ctx) {
      console.log('resolve', args.input.orgId);

      args.input.orgId.typename; // typescript validates

      return true;
    },
  },
  {
    outputFields: t => ({
      ok: t.boolean({ resolve: () => true }),
    }),
  },
);

```

Output with Relay first (matches TS types correctly):
```
authScopes { typename: 'Org', id: '123' }
resolve { typename: 'Org', id: '123' }
```

Output with Auth first (wrong):
```
authScopes T3JnOjEyMw==
resolve { typename: 'Org', id: '123' }

```

TLadd:
Fixes #708

Ran into this myself and took a crack at clarifying in the docs a bit.

### DIFF
--- a/packages/deno/packages/migration/mod.ts
+++ b/packages/deno/packages/migration/mod.ts
@@ -1,3 +1,0 @@
-import Default from './index.ts';
-export * from './index.ts';
-export default Default;
--- a/packages/deno/packages/plugin-add-graphql/mod.ts
+++ b/packages/deno/packages/plugin-add-graphql/mod.ts
@@ -1,3 +1,0 @@
-import Default from './index.ts';
-export * from './index.ts';
-export default Default;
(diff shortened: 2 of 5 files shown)


hayes (approved):
Thanks so much, this is great!

PR #794, opened 2023-02-20
feat: unwrap Promise fields for expose, fixes #793

### LINKED ISSUE #793: Allow expose to unwrap Promise types
Looking for something like this:

```ts
export class Example {
  get someStringField() {
    return Promise.resolve('value') // works if this is 'value'
  }
}

builder.objectType(Example, {
  name: 'Example',
  fields: (t) => ({
    someStringField: t.exposeString('someStringField'), // looking for this to work with string | Promise<string>
  }),
})
```

Haven't dug in to see if there's any reason this wouldn't be doable

tgriesser:
Changes `ParentShape[K]` to `Awaited<ParentShape[K]>` for determining if a field can be exposed via `.expose*` and for nullability

Closes #793

### DIFF
--- a/.changeset/fifty-cooks-cheer.md
+++ b/.changeset/fifty-cooks-cheer.md
@@ -1,0 +1,6 @@
+---
+'@pothos/core': minor
+'@pothos/deno': minor
+---
+
+Allow / unwrap Promises in "expose" type fields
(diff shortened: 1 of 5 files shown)


hayes (approved):
This looks good to me, thanks!

tgriesser:
Nice! Added a changeset, let me know if I did that correctly / feel free to change it if I didn't

hayes:
Yup, that's perfect.  Should be released soon

PR #797, opened 2023-02-21
bug: failing tests & fixes for id coercion

tgriesser:
Was working on adding `parse` to `loadableNode` - got pretty close to having it correct, but while working through the implementation I ran into a few bugs that seemed better to address separately before changing things around a bit for that.

The inferred types for `parse` are not properly reflecting the runtime coercion.

One of the issues was because `parse` was supposed to be `parseId` in the `relayGlobalIDFor`. The other issue is that it wasn't decoding the ids when passed to a field like `node`.

Not sure if this is 100% correct, the `parse` -> `parseId` seemed to just be an oversight, but wasn't sure if there was something else I was missing in the other changes.

### DIFF
--- a/packages/plugin-relay/src/index.ts
+++ b/packages/plugin-relay/src/index.ts
@@ -47,7 +47,7 @@ export class PothosRelayPlugin<Types extends SchemaTypes> extends BasePlugin<Typ
           String(globalID),
           ctx,
           info,
-          Array.isArray(mappings.value) ? mappings.value : false,
+          Array.isArray(mappings.value) ? mappings.value : true,
         ),
     );
 
--- a/packages/deno/packages/plugin-relay/field-builder.ts
+++ b/packages/deno/packages/plugin-relay/field-builder.ts
@@ -80,7 +80,7 @@ fieldBuilderProto.nodeList = function nodeList({ ids, ...options }) {
             const globalIds = rawIds.map((id) => typeof id === "string"
                 ? internalDecodeGlobalID(this.builder, id, context, info, true)
                 : id && {
-                    id: String(id.id),
+                    id: id.id,
                     typename: this.builder.configStore.getTypeConfig(id.type).name,
                 });
             return resolveNodes(this.builder, context, info, globalIds);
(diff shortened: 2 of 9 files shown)


packages/plugin-relay/src/index.ts
```diff
@@ -47,7 +47,7 @@ export class PothosRelayPlugin<Types extends SchemaTypes> extends BasePlugin<Typ
           String(globalID),
           ctx,
           info,
-          Array.isArray(mappings.value) ? mappings.value : false,
+          Array.isArray(mappings.value) ? mappings.value : true,
```
hayes, line 50: I think this will result in always parsing IDs, which would be ideal of we could keep the types in sync, unfortunately I don't think that's possible.  When you declare a globalID are without declaring the type it's for, it could be an ID for any type, and don't know how it's parsed.  If we do parse it, we would likely need to type it as unknown.  Keeping this case with IDs being parsed as strings seems like a better compromise.   I think we need to handle node(s) fields by parsing explicitly before calling the load methods rather than depending on the default arg parsing logic.  We could potentially add an argument to globalID fields to indicate they should always be parsed, or do something like for: '*'.

hayes:
Thanks for catching these issues!  I think we need to tweak the solution for the node(s) fields, but the other change looks good

packages/plugin-relay/src/index.ts
tgriesser: > When you declare a globalID are without declaring the type it’s for, it could be an ID for any type, and don’t know how it’s parsed.

Might be misunderstanding, but it seems like it's determining the typename based on unwrapping the ID, and then based on that typename if it has a parse function it runs it through that?

If you change this back to `false` you'll be able to see the tests that are failing that this was aiming to fix. 

> I think we need to handle node(s) fields by parsing explicitly before calling the load methods rather than depending on the default arg parsing logic. 

That sounds like it makes sense, maybe we can create a test case illustrating what you're referring to here so we ensure that's covered / doesn't regress?

hayes: I'm traveling for the next few weeks, but I should have a little time tonight to look at this.  I'll add some tests to cover those use cases

tgriesser: Thanks! No rush, free to delay until you're done traveling

PR #798, opened 2023-02-21
feat: add parse to loadableNode id

tgriesser:
Adds `parse` to the `id` field in `loadableNode`. Reorganizes the generics a bit, so we can assume we are receiving the parsed `ID` value as the `Key` for the DataLoader by default.

Dependent on the fixes in #797

~Edit: This isn't 100% correct, it infers the type of the value even when `parse` is not provided due to how `Key` is used. Need to dig into this a bit more~

Updated in https://github.com/hayes/pothos/pull/798/commits/9d9369285066063d2ef95e8c10d382b15752e329 to address ^, unclear if changes to the inferred generics like this constitutes a breaking change or not


packages/plugin-dataloader/tests/example/builder.ts
```diff
@@ -11,7 +11,7 @@ export default new SchemaBuilder<{ Context: ContextType }>({
     nodeQueryOptions: {},
     nodesQueryOptions: {},
   },
-  plugins: [RelayPlugin, ErrorsPlugin, DataloaderPlugin],
+  plugins: [ErrorsPlugin, DataloaderPlugin, RelayPlugin],
```
tgriesser, line 14: Not sure if we should enforce that `RelayPlugin` come before `DataLoaderPlugin` if we detect both are included, otherwise the IDs won't be parsed before loading.

packages/plugin-relay/src/input-field-builder.ts
```diff
@@ -37,7 +36,7 @@ inputFieldBuilder.globalIDList = function globalIDList<Req extends FieldRequired
             (Array.isArray(forTypes) ? forTypes : [forTypes])) as ObjectRef<SchemaTypes>[]
         )?.map((type: ObjectRef<SchemaTypes>) => ({
           typename: this.builder.configStore.getTypeConfig(type).name,
-          parse: type instanceof NodeRef ? type.parseId : undefined,
+          parseId: 'parseId' in type ? type.parseId : undefined,
```
tgriesser, line 39: Duck-type was the best I could come up with here, without moving `NodeRef` out of `plugin-relay` into `core` so it could be used by both plugins. Seems like it should be fine, since it shouldn't be confused with `parseId` anywhere else

packages/plugin-dataloader/tests/example/builder.ts
hayes: I think at some point having some sort of mechanism for ensuring plugin order would be good, but I don't want to address it in the PR.  Ideally the solution would be something generic other plugins can take advantage of.

PR #799, opened 2023-02-21
feat: handle string contining ':' in global ID

tgriesser:
It's an edge case, but if an ID were to have a `:` in it for some reason, this change ensures that it gets included rather than splitting and only including the first value after the typename `:` delimiter


hayes (approved):
Looks good!

PR #800, opened 2023-02-21
feat: implement default isTypeOf on loadableNode

tgriesser:
Digging more into `loadableNode`, realized it doesn't implement a default `isTypeOf` in the same way that `node` does.

In the docs, there's a comment:

`// For loadable objects we always need to include an isTypeOf check`

It seems like if `loadableNode` takes a ref, it should be possible to handle this with the default implementation Relay provides?

Anyway, this PR does that. Not sure if we should be reusing the logic between the two plugins (maybe we should move the default `isTypeOf` to `core`)?


hayes:
merged as #808

PR #801, opened 2023-02-22
feat: allow unionType types to be a thunk

tgriesser:
Didn't see any reason `unionType` couldn't be a thunk, the same way that is supported in `GraphQLUnionType`. 
Helpful in situations with circular references

### DIFF
--- a/packages/core/src/builder.ts
+++ b/packages/core/src/builder.ts
@@ -423,15 +423,17 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
       ParentShape<Types, Member>
     >(name);
 
-    options.types.forEach((type) => {
-      verifyRef(type);
-    });
+    if (Array.isArray(options.types)) {
+      options.types.forEach((type) => {
+        verifyRef(type);
+      });
+    }
 
     const config: PothosUnionTypeConfig = {
       kind: 'Union',
       graphqlKind: 'Union',
       name,
-      types: (options.types || []) as ObjectParam<SchemaTypes>[],
+      types: [],
       description: options.description,
       resolveType: options.resolveType as GraphQLTypeResolver<unknown, object>,
       pothosOptions: options as unknown as PothosSchemaTypes.UnionTypeOptions,
@@ -439,6 +441,7 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
     };
 
     this.configStore.addTypeConfig(config, ref);
+    this.configStore.addUnionTypes(name, options.types);
 
     return ref;
   }
--- a/.changeset/great-suits-provide.md
+++ b/.changeset/great-suits-provide.md
@@ -1,0 +1,6 @@
+---
+'@pothos/core': minor
+'@pothos/deno': minor
+---
+
+Allow unionType to receive types as a thunk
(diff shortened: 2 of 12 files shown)


packages/core/src/builder.ts
```diff
@@ -423,15 +423,13 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
       ParentShape<Types, Member>
     >(name);
 
-    options.types.forEach((type) => {
-      verifyRef(type);
```
hayes, line 427: I'd like to leave this here if this is not a function.  The issue this is usually detecting is a ref that's undefined because of a circular import issue (or something similar). Using a thunk should help, but by evaluating here when it's an array, you get a stacktrace that actually points to the right location.

packages/core/src/builder.ts
```diff
@@ -440,6 +438,13 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
 
     this.configStore.addTypeConfig(config, ref);
 
+    this.configStore.onTypeConfig(ref, () => {
+      const unionTypes = typeof types === 'function' ? types() : types;
+      unionTypes.forEach((type) => {
+        verifyRef(type);
```
hayes, line 444: I don't think that this will be necessary, as mentioned above, this is mostly to detect when refs are undefined which I've only seen happen because of import issues.  The lazy evaluation should help with that

hayes (commented):
Overall this is a great feature! It's a bit more work and adds some complexity, but if we could keep the think our of the config-store that would probably be better.  This would be a potential breaking change for the plugin API, and having to unwrap thinks in plugins would be inconvenient.  We already do something very similar for interfaces (https://github.com/hayes/pothos/blob/main/packages/core/src/config-store.ts#L101)

packages/core/src/builder.ts
tgriesser: Ah nice, will update

hayes (commented):
For this to not be a breaking change, I think we need to use a pattern similar to what is done for interfaces on an object in the config store.

Basically the Union should be added to the config with an empty array of members if it's a function, and then the thunk should be passed to a new addMembers method.  This should store the members thunks to be resolved when toSchema is called. See the addInterfaces method for how this could work: (https://github.com/hayes/pothos/blob/main/packages/core/src/config-store.ts#L101)

hayes:
@tgriesser do you have a circular reference use case where this was an issue?   I've wondered about this in the past, and couldn't come up with one.

Fields are already deferred, and unions can't contain other unions, which is why I haven't done this yet.  It seems like it definitely should be supported, but I hadn't actually run into a use case for it yet.

tgriesser:
> For this to not be a breaking change, I think we need to use a pattern similar to what is done for interfaces on an object in the config store.

Ah cool, will look into doing this - re-requested the review before I saw that comment

> do you have a circular reference use case where this was an issue? I've wondered about this in the past, and couldn't come up with one.

Yeah I just ran into an issue where where I could technically avoid it by creating a new file for the union, but it'd be simpler to just use a thunk.

```
/typeA.ts - imports typeB for use in a field, also has a union with TypeB (eagerly evaluated)
/typeB.ts - imports typeA for use in a field
```


I think it might have been further caused by using esbuild on the server here, and it might have slightly different loading semantics than normal modules? Not 100%

hayes (approved):
Awesome, thanks so much!

PR #819, opened 2023-02-28
In SubGraph plugin, use a Set to avoid duplicate interfaces

### LINKED ISSUE #818: Duplicate interfaces in SDL when using the SubGraph and Relay plugins
Noticed a little bug in the interaction of the SubGraph and Relay plugins:

When adding a Node type, the type in the second subGraph receives an extra Node interface:
`type Foo implements Node & Node`

Here is a minimal repro that prints 2 subgraph schemas:
https://github.com/hayes/pothos/commit/ad5d4853148dc1ddfa50593af8405c9dda503288

Type in the first schema is fine (normal `implements Node`)

davidgovea:
Attempts to fix #818

This --does-- repair the behavior of the repro at https://github.com/hayes/pothos/commit/ad5d4853148dc1ddfa50593af8405c9dda503288

Screen Shot 2023-02-28 at 11 41 14 AM

But, this is probably not the best way to do it. Seems like the problem is originating elsewhere in the interface assembly, and this is just an incomplete patch.

Also, the filter is applied to Object types only, but there is similar code for Interface types
https://github.com/hayes/pothos/blob/feafaa64ad4f9fe7827b45908745f883a4238279/packages/plugin-sub-graph/src/index.ts#L138-L141

---

Happy to run with any guidance you have here, investigate more, add tests etc :)

### DIFF
--- a/packages/plugin-sub-graph/src/index.ts
+++ b/packages/plugin-sub-graph/src/index.ts
@@ -122,10 +122,13 @@ export class PothosSubGraphPlugin<Types extends SchemaTypes> extends BasePlugin<
           type.name,
           new GraphQLObjectType({
             ...typeConfig,
-            interfaces: () =>
-              typeConfig.interfaces
-                .filter((iface) => newTypes.has(iface.name))
-                .map((iface) => replaceType(iface, newTypes, typeConfig.name, subGraphs)),
+            interfaces: () => [
+              ...new Set(
+                typeConfig.interfaces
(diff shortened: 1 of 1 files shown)


hayes:
Thanks for the fix!  I looked into this, and I'm pretty sure this is a bug in core where interfaces where getting re-added each time the schema was built.  I just pushed a fix that should hopefully fix the underlying issue

PR #825, opened 2023-03-06
feat: add bigint to DefaultScalars ID value

tgriesser:
Adds `bigint` to the DefaultScalars definition for the ID field, so the return value of `id.resolve` in the `node` / `loadableNode` helper can be a `BigInt`

image
image

### DIFF
--- a/packages/plugin-relay/package.json
+++ b/packages/plugin-relay/package.json
@@ -57,5 +57,8 @@
     "graphql-subscriptions": "^2.0.0",
     "graphql-tag": "^2.12.6"
   },
+  "engines": {
+    "node": ">=12.0.0"
+  },
   "gitHead": "9dfe52f1975f41a111e01bf96a20033a914e2acc"
 }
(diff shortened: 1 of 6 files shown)


packages/plugin-relay/package.json
```diff
@@ -57,5 +57,8 @@
     "graphql-subscriptions": "^2.0.0",
     "graphql-tag": "^2.12.6"
   },
+  "engines": {
+    "node": ">=12.0.0"
```
tgriesser, line 61: Added this so ESLint doesn't complain that `BigInt` doesn't exist yet, by default eslint checks down to 8.0.0 if no engines are provided.

hayes:
@tgriesser did you mean to close this?

I think adding bigint as an allowed default export would probably be okay.  Looks like you change changed both the input and output types which you probably don't want.

The current way to handle this is just to customize the scalar type when setting up the builder (this is how I set up for most apps I maintain):

```ts
  Scalars: {
    ID: {
      Input: string
      Output: number | string | bigint;
   }
}>(...)
```

ID inputs changing to just string is already on the 4.0 todolist as a breaking change

tgriesser:
I ended up realizing that in general it’s easier to just keep bigints as strings when using as IDs rather than casting them to actual BigInts. That makes sense though, I was wondering about the input vs output values

PR #827, opened 2023-03-07
add nullability to prismaNode

MaxAst:


packages/plugin-prisma/src/schema-builder.ts
```diff
@@ -105,14 +106,14 @@ schemaBuilderProto.prismaNode = function prismaNode(
...
-      const record = await (delegate.findUniqueOrThrow
+      const record = await (delegate.findUniqueOrThrow && !nullable
         ? delegate.findUniqueOrThrow({
             ...query,
             where: rawFindUnique ? rawFindUnique(id, context) : { [fieldName]: idParser!(id) },
           } as never)
         : delegate.findUnique({
             ...query,
-            rejectOnNotFound: true,
+            rejectOnNotFound: nullable ? false : true,
```
hayes, line 116: `...(nullable ? {} : { rejectOnNotFound: true },`

hayes:
Looks good, would be great to get something added to the docs and add a changeset. 

Thanks for adding this feature!

packages/plugin-prisma/tests/nodes.test.ts
```diff
@@ -90,7 +90,66 @@ describe('prisma', () => {
     `);
   });
 
-  it('query for single node', async () => {
+  it.only('query for nullable single node', async () => {
```
hayes, line 93: need to revert this `it.only`

hayes:
@MaxAst looks like some tests are failing, I haven't had time to dig into it yet, but I think it's other tests using the prismaNode you modified (there are lots of tests that assert the specific prisma calls being made, and it looks like the failing tests are seeing findUnique instead of findUniqueOrThrow)

MaxAst:
> @MaxAst looks like some tests are failing, I haven't had time to dig into it yet, but I think it's other tests using the prismaNode you modified (there are lots of tests that assert the specific prisma calls being made, and it looks like the failing tests are seeing findUnique instead of findUniqueOrThrow), which likely just need to be updated

@hayes my bad! just saw you fixed them already, thank you!

PR #830, opened 2023-03-09
[Fix]: Add defaultValue for args and input fields

moonflare:
### Context
We are migrating from Federation v1 spec to Federation v2, and for that, we need to change from using `printSchema` from `graphql` to `printSubgraphSchema` from `@apollo/subgraph`. Unfortunately, `printSubgraphSchema` is relying on the `astNode` to get the `defaultValue`s for inputs and arguments, so when printing the schema to the disk, we are missing those defaults from the schema. This is causing breaking changes when pushing the new schema (now with Fed v2+) to the registry.

### Description
* This PR makes sure we add the `defaultValue` to the `astNode` for both the arguments and the input fields.

### References
https://github.com/apollographql/federation/issues/2438

### DIFF
--- a/.changeset/nasty-trainers-remember.md
+++ b/.changeset/nasty-trainers-remember.md
@@ -1,0 +1,5 @@
+---
+"@pothos/plugin-directives": patch
+---
+
+Add ast nodes for defaultalues for args and input fields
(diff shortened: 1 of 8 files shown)


hayes (approved):
This looks great, Thanks so much!

Just need to add a changeset, and this should be good to go

PR #834, opened 2023-03-10
feat: allow connection fields to be Promises

tgriesser:
- Changes the types to permit the `edges` / `pageInfo` to be `MaybePromise<T>`
- Adds `completeValue` as a utility fn to remove the overhead of `async`/ await `await` on values that are `MaybePromise<T>`
  - Context for why this is helpful (not as big of an issue here, but a a small perf win): https://twitter.com/tgriesser/status/1212593239410262016

### DIFF
--- a/.changeset/plenty-zoos-act.md
+++ b/.changeset/plenty-zoos-act.md
@@ -5,4 +5,5 @@
 '@pothos/core': minor
 ---
 
-allow connection fields to be promises, add builder.completeValue helper
+- allow connection fields (edges / pageInfo) to be promises
+- add completeValue helper to core for unwrapping MaybePromise values
--- a/.changeset/plenty-zoos-act.md
+++ b/.changeset/plenty-zoos-act.md
@@ -7,3 +7,4 @@
 
 - allow connection fields (edges / pageInfo) to be promises
 - add completeValue helper to core for unwrapping MaybePromise values
+- set nodes as null if edges is null and the field permits a null return
--- a/.changeset/plenty-zoos-act.md
+++ b/.changeset/plenty-zoos-act.md
@@ -1,0 +1,8 @@
+---
+'@pothos/plugin-dataloader': minor
+'@pothos/plugin-prisma': minor
+'@pothos/plugin-relay': minor
+'@pothos/core': minor
+---
+
+allow connection fields to be promises, add builder.completeValue helper
--- a/packages/core/src/builder.ts
+++ b/packages/core/src/builder.ts
@@ -72,7 +72,13 @@ import type {
   SubscriptionFieldThunk,
   ValuesFromEnum,
 } from './types';
-import { normalizeEnumValues, valuesFromEnum, verifyInterfaces, verifyRef } from './utils';
+import {
+  isPromiseLike,
+  normalizeEnumValues,
+  valuesFromEnum,
+  verifyInterfaces,
+  verifyRef,
+} from './utils';
 
 export default class SchemaBuilder<Types extends SchemaTypes> {
   static plugins: Partial<PluginConstructorMap<SchemaTypes>> = {};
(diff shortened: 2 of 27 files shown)


packages/core/src/utils/index.ts
```diff
@@ -138,3 +138,9 @@ export function unwrapInputListParam<Types extends SchemaTypes>(
 
   return param;
 }
+
+export function isPromiseLike<T>(value: unknown): value is PromiseLike<T> {
```
hayes, line 142: I think there is an existing helper (isThenable) that does the same thing

packages/plugin-relay/src/schema-builder.ts
```diff
@@ -287,16 +288,17 @@ schemaBuilderProto.node = function node(param, { interfaces, extensions, id, ...
...
         ...this.options.relayOptions.idFieldOptions,
         ...id,
         args: {},
-        resolve: async (parent, args, context, info) => ({
-          type: nodeConfig.name,
+        resolve: (parent, args, context, info) =>
           // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-          id: await id.resolve(parent, args, context, info),
-        }),
+          this.completeValue(id.resolve(parent, args, context, info), (globalId) => ({
```
hayes, line 298: Love this change.  I had done this for a few performance critical places in the past and was really surprised by how much of an improvement this type of change makes!

packages/core/src/builder.ts
```diff
@@ -611,4 +617,46 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
...
+
+  /**
+   * Helper for allowing plugins to fulfill the return of the `next` resolver, without paying the cost of the
+   * Promise if not required.
+   */
+  completeValue<T, R, E>(
+    valOrPromise: PromiseLike<T> | T,
+    onSuccess: (completedVal: T) => R,
+    onError: (err: unknown) => E,
+  ): R | Awaited<R> | Promise<Awaited<R>> | Promise<Awaited<R>>;
```
hayes, line 638: Duplicate in the return type

hayes: Also is Awaited<R> an actual possible return type?

packages/core/src/builder.ts
```diff
@@ -611,4 +617,46 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
...
+
+  /**
+   * Helper for allowing plugins to fulfill the return of the `next` resolver, without paying the cost of the
+   * Promise if not required.
+   */
+  completeValue<T, R>(
```
hayes, line 644: Any reason this needs to be on the builder rather than exported along with other utils?

hayes (commented):
Thanks for adding this! Looks like a great change

packages/core/src/builder.ts
tgriesser: Yeah was just changing this to be a util fn

packages/core/src/builder.ts
```diff
@@ -611,4 +617,46 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
...
+    if (isPromiseLike(valOrPromise)) {
+      return valOrPromise.then(onSuccess, onError);
+    }
+    // No need to handle onError, this should just be a try/catch inside the `onSuccess` block
+    const result = onSuccess(valOrPromise);
+
+    // If the result of the synchronous call is a promise, we want to unwrap it, for
+    // the return value types consistency
+    if (isPromiseLike(result)) {
+      return result.then((o) => o);
```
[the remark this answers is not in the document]
tgriesser: Updated to just make it `Promise.resolve(result)`

It makes more sense from where I ripped it out of Nexus, where there's a lot more `PromiseLike` usage supported - basically the idea is that if you have a `result.then` which is a thenable and not a Promise, it unwraps the then call to get the returned value of the `.then` for type purposes without fully coercing it to a Promise.

packages/core/src/utils/index.ts
tgriesser: Updated to use that, and changed `isThenable` to return `PromiseLike`

packages/core/src/builder.ts
tgriesser: I think the `Awaited` was only needed as a return type when supporting the `PromiseLike` unwrapping thing I mentioned you had a question about above. It isn't a realistic scenario to support I don't think

packages/plugin-relay/tests/index.test.ts
```diff
@@ -484,6 +484,10 @@ describe('relay example schema', () => {
         contextValue: {},
       });
 
+      if (result.errors?.length) {
+        console.log(result.errors.map((e) => e.stack));
+      }
+
```
tgriesser, line 490:
```suggestion
```

.changeset/plenty-zoos-act.md
```diff
@@ -0,0 +1,8 @@
+---
+'@pothos/plugin-dataloader': minor
+'@pothos/plugin-prisma': minor
+'@pothos/plugin-relay': minor
+'@pothos/core': minor
+---
+
+allow connection fields to be promises, add builder.completeValue helper
```
hayes, line 8: This probably should be updated too

packages/plugin-relay/src/schema-builder.ts
```diff
@@ -521,7 +524,11 @@ schemaBuilderProto.connectionObject = function connectionObject(
                   this.options.relayOptions?.nodeFieldOptions?.nullable ??
                   false,
               },
-              resolve: (con) => (con.edges?.map((edge) => edge?.node) ?? []) as never,
+              resolve: (con) =>
+                completeValue(
+                  con.edges,
+                  (edges) => edges?.map((e) => e?.node) ?? (edgeListNullable ? null : []),
```
tgriesser, line 530: @hayes not sure if this change needs mentioning, but previously `nodes` always `returned []` if edges was null, even if the field itsself was nullable.

tgriesser: Just added a note

hayes:
Looks like type checking is failing, you can rub pnpm ci to run the full set of checks that get run against PRs

packages/plugin-relay/src/utils/connections.ts
```diff
@@ -3,6 +3,7 @@ import {
   encodeBase64,
   MaybePromise,
   PothosValidationError,
+  RemoveAwaitedProps,
```
tgriesser, line 6: Since we know the return types of the connection helpers, we can remove the `MaybePromise` wrappers on the return type.

Issue #856, opened 2023-03-28
Subgraph Plugin is dropping deprecation hints in InputTypes


boenni23:
When you use `builder.toSchema`  with `subGraph` you will loose the `@deprecated` directives in the resulting schema.

Taking the example from 
https://pothos-graphql.dev/docs/plugins/sub-graph#subgraph-plugin
the resulting schema will preserve `@deprecated` in **types** but NOT in **inputs** or **args**.

Having this example:

```

builder.inputType('PostWhereInput', {
  subGraphs: ['Internal'],
  fields: (t): any => ({
    id: t.field({
      type: 'ID',
      required: false,
    }),
    title: t.field({
      type: 'String',
      required: false,
      deprecationReason: 'Use `id` instead',
    }),
  }),
})

builder.objectType('Post', {
  subGraphs: ['Internal'],
  fields: (t): any => ({
    id: t.exposeID('id'),
    title: t.exposeString('title', {
      deprecationReason: 'will not be used anymore',
    }),
  }),
})

builder.queryField('posts', (t) =>
  t.field({
    type: ['Post'],
    args: {
      where: t.arg({
        type: 'PostWhereInput',
        deprecationReason: 'Will be someting else',
        required: false,
      }),
    },
    resolve: (root, args, ctx, info) => {
      return []
    },
  }),
)

```

Comparing `builder.toSchema({})` with `builder.toSchema({ subGraph: 'Internal' })` Schemas via 
```
const saveSchema = (schema: GraphQLSchema, fileName: string) =>
  writeFileSync(fileName, printSchema(lexicographicSortSchema(schema)))
```

```diff
type Post {
  id: ID!
  title: String! @deprecated(reason: "will not be used anymore")
}

input PostWhereInput {
  id: ID
-  title: String @deprecated(reason: "Use `id` instead")
+  title: String
}

type Query {
-  posts(where: PostWhereInput @deprecated(reason: "Will be someting else")): [Post!]!
+  posts(where: PostWhereInput): [Post!]!
}
```

hayes:
@boenni23 published your fix, let me know if you are still having issues after upgrading.  Thanks for the report and the fix!

PR #857, opened 2023-03-28
In SubGraph Plugin: deprecation hints in InputTypes and Args

boenni23:
Partial Fix for #856 

Sorry but I could not fix `deprecated` for query input args. Just did not find it.


hayes:
Thanks for this fix!  I added some tests, and it looks like your fix already applied to query arguments and inputs.  Were you testing something specific that you were testing that wasn't working with this fix?

boenni23:
You are right. No more issues with in the schema. Thanks

PR #880, opened 2023-04-19
Add ID arg options for `node` and `nodes` queries

psirenny:
This allows arg options to be set for the ID arguments in `node` and `nodes` queries. My motivation was to be able to provide descriptions for each argument. For example:

```ts
import SchemaBuilder from "@pothos/core";
import RelayPlugin from "@pothos/plugin-relay";

const schemaBuilder = new SchemaBuilder<{}>({
  plugins: [RelayPlugin],
  relayOptions: {
    nodeQueryIdArgOptions: {
      description: "The ID of the node.",
    },
    nodesQueryIdsArgOptions: {
      description: "The IDs of each node.",
    },
  },
});
```

### DIFF
--- a/packages/plugin-relay/src/types.ts
+++ b/packages/plugin-relay/src/types.ts
@@ -91,7 +91,7 @@ export type RelayPluginOptions<Types extends SchemaTypes> = EmptyToOptional<{
           resolveNodes: (ids: { id: string; typename: string }[]) => Promise<unknown[]>,
         ) => MaybePromise<readonly MaybePromise<unknown>[]>;
       });
-  nodesQueryIdArgOptions: Omit<
+  nodesQueryIdsArgOptions: Omit<
     GlobalIDListInputFieldOptions<Types, true, 'Arg', ObjectParam<Types>>,
     'extensions' | 'required'
   >;
--- a/packages/plugin-relay/src/types.ts
+++ b/packages/plugin-relay/src/types.ts
@@ -55,6 +55,9 @@ export type RelayPluginOptions<Types extends SchemaTypes> = EmptyToOptional<{
         >,
         'args' | 'resolve' | 'type'
       > & {
+        args?: {
+          id?: Omit<GlobalIDInputFieldOptions<Types, true, 'Arg', ObjectParam<Types>>, 'required'>;
+        };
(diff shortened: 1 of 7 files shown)


packages/plugin-relay/src/types.ts
```diff
@@ -65,6 +65,10 @@ export type RelayPluginOptions<Types extends SchemaTypes> = EmptyToOptional<{
           resolveNode: (id: { id: string; typename: string }) => MaybePromise<unknown>,
         ) => MaybePromise<unknown>;
       });
+  nodeQueryIdArgOptions: Omit<
+    GlobalIDInputFieldOptions<Types, true, 'Arg', ObjectParam<Types>>,
+    'extensions' | 'required'
```
hayes, line 70: I think it would be better to allow extensions, and just merge them in

hayes (approved):
Thanks for the contribution! definitely want as much of this to be customizable as possible.

I am wondering if it would be better to nest the id arg options inside the options for the query field instead like:

```ts
nodesQueryOptions: { idArg: { description } }
```

Or something similar.  Maybe that would not be very discoverable.  The top level already feels very cluttered, but maybe that's okay.

packages/plugin-relay/src/types.ts
psirenny: Agreed, the extensions are merged in now as well 😄.

psirenny:
> I am wondering if it would be better to nest the id arg options inside the options for the query field instead like:
> 
> ```ts
> nodesQueryOptions: { idArg: { description } }
> ```
> 
> Or something similar. Maybe that would not be very discoverable. The top level already feels very cluttered, but maybe that's okay.

Yea, when I was looking to add argument descriptions, I first checked the `nodeQueryOptions` and `nodesQueryOptions` to see if they allowed for that customization. It's a natural place to search for those options so it might improve discoverability. Although, I wasn't sure how best to document it in the README.

I went for something like `nodesQueryOptions: { args: { id: { description } } }` because the destructuring logic was cleaner. I also figure that pattern could scale to additional arguments on other fields if that's needed at any point.

hayes:
This looks good, thanks again!

PR #886, opened 2023-04-25
Include `idFieldOptions` in the Node interface ID field

psirenny:
This will also apply `idFieldOptions` to the Node interface. My goal was to give the Node interface's ID field a description:

```ts
const schemaBuilder = new SchemaBuilder({
  plugins: [RelayPlugin],
  relayOptions: {
    idFieldOptions: {
      description: "The node’s globally unique identifier.",
    },
  },
};
```

```graphql
interface Node {
  "The node’s globally unique identifier."
  id: ID!
}
```

### DIFF
--- a/.changeset/pink-falcons-rule.md
+++ b/.changeset/pink-falcons-rule.md
@@ -1,0 +1,6 @@
+---
+'@pothos/plugin-relay': minor
+'@pothos/deno': minor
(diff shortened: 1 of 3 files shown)


hayes (approved):
This makes sense, thanks for the PR!

PR #890, opened 2023-04-30
Add ability to configure meta for builder.enumType(<ts enum>)

kelleyvanevert:
Hey!

I'm working on a pull request to add the ability to also optionally specify graphql metadata (description, deprecationReason, and extensions) for an enum type, even if the enum type is provided as a TypeScript enum. Because currently, you cannot do this:

```ts
enum SomeTypescriptEnum {
  bla = "bla",
  yeah = "yeah",
}

builder.enumType(SomeTypescriptEnum, {
  name: "SomeTypescriptEnum",

  // not currently possible:
  values: {
    bla: {
      deprecationReason: "We don't use this anymore for some reason",
    },
  }.
});
```

The actual JS fix is simple, but the TypeScript typings are a bit more complicated, and then it'll also require some changes in the other packages and docs I think, before this PR can be complete.

The first commit I just added allows the core package to build successfully, although I'm not 100% sure that the typings are as they're supposed to be.

If anyone wants to pitch in to help with the additional work, I'd be super happy! :)

Also, @hayes, I'm not 100% sure how all the tooling of your repo works, e.g. the changesets etc. So maybe you could help me out here :)

### DIFF
--- a/packages/core/src/builder.ts
+++ b/packages/core/src/builder.ts
@@ -458,8 +458,11 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
 
     const values =
       typeof param === 'object'
-        ? // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-          valuesFromEnum<Types>(param as BaseEnum, options?.values as any)
+        ? valuesFromEnum<Types>(
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+            param as BaseEnum,
+            options?.values as Record<string, PothosSchemaTypes.EnumValueConfig<Types>>,
+          )
         : normalizeEnumValues<Types>((options as { values: EnumValues<Types> }).values);
 
     const config: PothosEnumTypeConfig = {
--- a/packages/core/src/builder.ts
+++ b/packages/core/src/builder.ts
@@ -459,7 +459,7 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
     const values =
       typeof param === 'object'
         ? // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-          valuesFromEnum<Types>(param as BaseEnum)
+          valuesFromEnum<Types>(param as BaseEnum, options?.values as any)
         : normalizeEnumValues<Types>((options as { values: EnumValues<Types> }).values);
 
     const config: PothosEnumTypeConfig = {
--- a/packages/core/src/types/builder-options.ts
+++ b/packages/core/src/types/builder-options.ts
@@ -180,6 +180,9 @@ export type EnumTypeOptions<
   ? Merge<
       Omit<PothosSchemaTypes.EnumTypeOptions<Types, Values>, 'values'> & {
         name: string;
+        values?: Partial<
+          Record<keyof Param, Omit<PothosSchemaTypes.EnumValueConfig<Types>, 'value'>>
+        >;
       }
     >
   : PothosSchemaTypes.EnumTypeOptions<Types, Values>;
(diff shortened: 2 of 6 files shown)


hayes:
changesets are pretty simple, just run `pnpm changeset` from the root of the repo, and it will prompt you through the processes.  It will ask you what packages are you changed, if its a patch, minor, or major, and then ask for a description of the change for the changelog.  This will create a new changeset file you can add to the PR.

packages/core/src/types/builder-options.ts
```diff
@@ -175,14 +180,14 @@ export type InterfaceTypeOptions<
 export type EnumTypeOptions<
   Types extends SchemaTypes,
   Param extends EnumParam,
-  Values extends EnumValues<Types>,
+  Values extends EnumValues<Types, Param>,
```
hayes, line 183: this would be a breaking change, and will break other packages (I think you already mentioned this in the PR description)

packages/core/src/types/builder-options.ts
```diff
@@ -175,14 +180,14 @@ export type InterfaceTypeOptions<
...
 > = Param extends BaseEnum
   ? Merge<
-      Omit<PothosSchemaTypes.EnumTypeOptions<Types, Values>, 'values'> & {
+      PartialBy<PothosSchemaTypes.EnumTypeOptions<Types, Param, Values>, 'values'> & {
```
hayes, line 186: rather than trying to hack around this special case, it will be a lot easier just to leave all the existing type as is, and add in values on the line below.

Basically, you should be able to add a new field like `values?: Partial<Record<KeysFromParam<Param>, Omit<EnumValueOptions, 'value'>>`  on the line below, along side name.  That object already is customizing the options for the case where the Param was an Enum value, and the normal values option has been removed, so you can add a new values option specific to this case.

The type in the comment above is awful sudo-code and not actually how that would be implemented, but something similar should work.

hayes: doing it this way should avoid needing to change any of the existing shared types, so that this doesn't break existing usages of those types.

hayes:
thanks for opening this PR!  This seems like a great feature, happy to answer any questions to help you get this implemented properly.  Always happy to have new contributors!

kelleyvanevert:
Good idea @hayes, I've implemented it now in the way you suggested, as a minor non-breaking change not changing the existing types, and included a changeset line.

I think that the PR should then be complete now, maybe you can double-check? :)

packages/core/src/utils/enums.ts
```diff
@@ -32,6 +32,7 @@ export function normalizeEnumValues<Types extends SchemaTypes>(
 
 export function valuesFromEnum<Types extends SchemaTypes>(
   Enum: BaseEnum,
+  values?: Partial<Record<string, Omit<PothosSchemaTypes.EnumValueConfig<Types>, 'value'>>>,
```
hayes, line 35: Partial doesn't look like it would do anything here

packages/core/src/builder.ts
```diff
@@ -459,7 +459,7 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
     const values =
       typeof param === 'object'
         ? // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-          valuesFromEnum<Types>(param as BaseEnum)
+          valuesFromEnum<Types>(param as BaseEnum, options?.values as any)
```
hayes, line 462: We general try to avoid any where ever possible in Pothos, casting to the type expected by the function should work (not 100% sure, there might be a comparability issue), you could cast through unknown, or cast to never if you really need to, but linting (pnpm lint) won't pass with the any

kelleyvanevert:
Awesome!

Thanks @hayes for helping out to land this small feature so fast! :)

PR #903, opened 2023-05-07
stop using unavailable prisma CreateInput objects

lsmacedo:
### Problem

When you have a Prisma model where one of the required fields are of an unsupported type, Prisma does not generate a `prisma.modelName.create` function, so rows of this type can only be created through raw queries. A problem behind this is that the `generated.ts` file gets an error trying to reference a non-existent type from Prisma, and we're not able to compile with `tsc` (even with skipLibCheck).

### Solution

This PR makes necessary adjustments on Pothos' prisma plugin generator in order to not try to use the `Prisma.${modelName}CreateInput` when it is not available.

### DIFF
--- a/.changeset/few-turtles-tell.md
+++ b/.changeset/few-turtles-tell.md
@@ -1,0 +1,5 @@
+---
+"@pothos/plugin-prisma": patch
+---
+
+Prevent unavailable prisma CreateInput types from being referenced by pothos generated types.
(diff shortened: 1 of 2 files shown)


hayes (approved):
Looks good, thanks for the fix!  If you add a changeset I can get this released for you

lsmacedo:
> Looks good, thanks for the fix! If you add a changeset I can get this released for you

Awesome. It is done.

Issue #910, opened 2023-05-16
Allow scalars option for SchemaBuilder


Liam-Tait:
Instead of adding scalars after creating a schema builder using `builder.addScalarType`, I would like to be able to give `SchemaBuilder` an object with scalars.

This would set up like the following
```ts
const builder = new SchemaBuilder({
  scalars: {
    'ScalarName': Resolver
  }
})
```

This would help with [this issue on using graphql-scalars](https://github.com/hayes/pothos/issues/726) and [this discussion for custom scalar methods](https://github.com/hayes/pothos/discussions/568)

As the resolvers are set up, methods can then be added for convenience.

This would look like the following with [graphql-scalars](http://graphql-scalars.dev/) and method generation. I have used the SimpleObjectsPlugin for simplicity

```ts
import SchemaBuilder from "@pothos/core"
import { EmailAddressResolver, DateTimeResolver } from "graphql-scalars"

const builder = new SchemaBuilder({
  plugins: [SimpleObjectsPlugin],
  scalars: {
    'DateTime': DateTimeResolver,
    'Email': EmailAddressResolver
  }
})

builder.simpleObject('Appointment', {
  fields: (t) => ({
    email: t.email()
    date: t.datetime(),
  })
})
```

This is useful as the types can be inferred from the resolver for `Input` and `Output`. This is also useful for not adding the extra methods as the field would still be correctly typed
e.g 
```ts
builder.simpleObject('Appointment', {
  fields: (t) => ({
    email: t.field({type: 'Email'})
    date: t.field({type: 'DateTime'}),
  })
})
```

hayes:
While this might be technically possible, there are a couple of issues:

* Input and Output types can't actually be inferred because the GraphQL classes don't actually support the right generics to make this work
* Dynamically inferring builder methods based on options gets really complicated and would require a major redesign of how options are types (breaking change) and would make the whole builder type a lot more complex, and would likely break the ability to create helper methods that accept builder instances

Liam-Tait:
> Input and Output types can't actually be inferred because the GraphQL classes don't actually support the right generics to make this work

My understanding is these would match `TInternal` and `TExternal` from `GraphQLScalarType`. Is that correct?

> Dynamically inferring builder methods based on options gets really complicated and would require a major redesign of how options are types (breaking change) and would make the whole builder type a lot more complex, and would likely break the ability to create helper methods that accept builder instances

Makes sense. I saw this as a happy side effect made possible by having the types up front anyway.

[...]

PR #914, opened 2023-05-16
add withScalar method

Liam-Tait:
Based on this issue https://github.com/hayes/pothos/issues/910

Add method `withScalars` to `SchemaBuilder` to enable loading scalars and inferring their types for convenience. This especially helps with using existing libraries like [graphql-scalars](http://graphql-scalars.dev/)

Loading scalars is done via a method instead of as an option to enable the inference to merge the Scalars from the manually typed scalars and the inferred scalars. This is because typescript doesn't have a way to do partial inference.

Previously
```ts
  Scalars: {
    Date: {
      Input: Date;
      Output: Date;
    };
  };
}>({});

builder.addScalarType('Date', CustomDateScalar, {});
```

Using withScalars
```ts
const builder = new SchemaBuilder({}).withScalars({ Date: CustomDateScalar });
```


Liam-Tait:
@hayes This is ready except for the documentation. How do you want to prioritise the `withScalars` approach vs manual typed in the documentation?

packages/core/src/builder.ts
```diff
@@ -614,4 +615,30 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
...
+                  Input: TInternal;
+                  Output: TExternal;
+                }
+              : never;
+          };
+        }
+      >
+    >;
+
+    for (const [name, scalar] of toPairs(scalars)) {
```
hayes, line 639: Feel free to use Object.entries here with some type casts, not sure if adding toPairs and dealing with license stuff is worth it (up to you).

packages/core/src/builder.ts
```diff
@@ -614,4 +615,30 @@ export default class SchemaBuilder<Types extends SchemaTypes> {
...
+      that.addScalarType(name, scalar, {});
```
hayes, line 640: I know it's inconvenient and a bit awkward and inconvenient, but it's pretty important to be able to provide these options. 

This could be `Date: [DateResolver, options]` or `Date: { type/resolver/implementation: DateResolver, ...otherOptions }`

Personally I like the second option (although the resolver terminology from graphql-scalars is not my favorite.

I'd probably make it work so you can pass either just the scalar, or scalar + options.  Inference will be slightly more complex but shouldn't be too bad.

packages/core/tests/scalars.test.ts
```diff
@@ -0,0 +1,312 @@
...
+        diet: t.expose('diet', {
+          exampleRequiredOptionFromPlugin: true,
+          type: Diet,
+        }),
+      }),
+    });
+
+    builder.objectType(Kiwi, {
+      name: 'Kiwi',
+      interfaces: ['Animal'],
```
hayes, line 136: Just noticed you are based in New Zealand, which is where I'm currently on vacation 🥝

packages/core/tests/scalars.test.ts
```diff
@@ -0,0 +1,312 @@
...
+      isTypeOf: (value) => value instanceof Kiwi,
+      description: 'Long necks, cool patterns, taller than you.',
```
hayes, line 138: Long beaks? Rounder than you

hayes (changes requested):
Thanks so much for working on this change!

Added a few notes and questions, but generally looks like the right direction!

Regarding documentation, I think I'd prefer to keep recommending the current approach as the default until: this adds helpers for defining custom scalar fields, and graphql-scalars publishes a version where the scalars all have types


Sorry it's taking me a while to get this review done, and let me know if anything is unclear.  Looking at this on my phone with spotty service, so probably not the best conditions for a good review

Liam-Tait:
[The pr to expose TInternal and TExternal from graphql-scalars](https://github.com/Urigo/graphql-scalars/issues/1919) has been merged, so should be in the next release of graphql-scalars

packages/core/src/builder.ts
Liam-Tait: I'm working on this, might take a bit of time

I am aiming to have the following be valid.
```ts
const builder = new SchemaBuilder({}).withScalars({
  Date: DateScalar,
  UUID: { scalar: UUIDScalar },
  PositiveInt: {
    scalar: PositiveIntScalar,
    options: {
      extensions: { codegenScalarType: "number" },
    },
  },
});
```

Does that look good to you?

packages/core/tests/scalars.test.ts
Liam-Tait: New Zealand is an wonderful country to be in. I hope you're having a fantastic time exploring and experiencing everything it has to offer. Enjoy your stay! 🥝

hayes:
@Liam-Tait been a busy few weeks for me, sorry I haven't checked in here in a bit.  Do you want me to clean this up and get it merged?

Liam-Tait:
@hayes no rush at all. Happy for you to pick this up if you would like to. I believe changes required are adding options to each scalar.

My company works in cycles, I’ll have time to continue work on this after 7th July when  we are in a “cooldown” mode if you are happy to wait that long

hayes:
If I end up having some time, I'll try to work on this, but no rush on my end either, might just leave it for for you!  Summers tend to get pretty busy for me, and I have less time outside work for hacking on side projects.

Liam-Tait:
I have updated this pr to have many chained calls to `withScalar(name, scalar, options)`

For example using many graphql-scalars
```ts

import {
  GraphQLEmailAddress,
  GraphQLHexColorCode,
  GraphQLBoolean,
  GraphQLISBN,
  GraphQLBigInt,
  GraphQLRGB,
  GraphQLDate,
  GraphQLJSONObject
} from 'graphql-scalars';
import SchemaBuilder from '@pothos/core';

const builder = new SchemaBuilder({})
  .withScalar('Email', GraphQLEmailAddress)
  .withScalar('HexColor', GraphQLHexColorCode)
  .withScalar('Boolean', GraphQLBoolean)
  .withScalar('ISBN', GraphQLISBN)
  .withScalar("BigInt", GraphQLBigInt)
  .withScalar("RGB", GraphQLRGB)
  .withScalar("Date", GraphQLDate)
  .withScalar("JSONObject", GraphQLJSONObject)
```

Liam-Tait:
@hayes Are you happy with the chaining `withScalar` instead of `withScalars` taking an object?

hayes:
Hey, sorry I missed your last update!  Chaining APIs create a lot of complexity in the type-system, and I am not sure how this would affect the builder types.   Early versions of Pothos experimented with using chaining APIs to define fields, the issue was that you could very easily hit max stack depth limits in the type-system because each chained API is dependent on the result of the previous call.   Similarly, this API would make every builder method dependent on a chain of type-transforms.  I am not sure if this would actually cause issues in the real world, but is definitely something I would be concerned about, and would need to explore in depth before adding an API like that.

Liam-Tait:
Closing the pr as it has started to have merge conflicts, even if they are small.
I'm happy to fix up the pr, if it is likely to be merged.

PR #919, opened 2023-05-24
Add Documentation about Usage with `graphql-scalars`

### LINKED ISSUE #726: Can we use `graphql-scalars` scalars?
It would be nice to be able to add custom scalars using the [`graphql-scalars`](https://the-guild.dev/graphql/scalars/docs/scalars/big-int) package to the pothos builder.

Is there a way to do this?

arimgibson:
Closes #726.

Not sure if this is still relevant because of #910 but already had it 90% finished and kept forgetting to check formatting and PR... woops.

Anyway, feel free to play around with wording or whatever else if it's still relevant! Only concern is that the `JSONResolver` gets highlighted a little weirdly by highlight.js, but I'm not sure if there's a way to fix this.

image

### DIFF
--- a/website/pages/docs/guide/scalars.mdx
+++ b/website/pages/docs/guide/scalars.mdx
@@ -15,7 +15,7 @@ export const getStaticProps = () => ({ props: { nav: buildNav() }})
 
 # Scalars
 
-## Adding GraphQL Scalars
+## Adding Custom GraphQL Scalars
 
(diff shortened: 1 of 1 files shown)


hayes (approved):
Thanks so much for adding this!

hayes:
Edited the wording a bit to be less opinionated, but overall love this change, thanks!

PR #936, opened 2023-06-15
plugin-complexity: fix bug when limits are zero

arthursimas1:
hi all!

thanks for that awesome library. it's way easier to implement a graphql api using it!

while experimenting with it, i noticed that the plugin `@pothos/plugin-complexity` interprets zero-values as just an undefined limit.

i expect that a zero-value means that a limit is reached, just like a negative value.

i've changed the truthy comparison to a `typeof <limit name> === 'number'` where applicable. i've also added some test cases that cover zero-values and negatives (already supported, but just in case...).

i'd be glad if you consider incorporating this change.

### DIFF
--- a/.changeset/wild-swans-heal.md
+++ b/.changeset/wild-swans-heal.md
@@ -1,0 +1,5 @@
+---
+'@pothos/plugin-complexity': minor
+---
+
+fix bug when limits are zero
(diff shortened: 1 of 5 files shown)


arthursimas1:
probably a version bump should be needed, as it may break things i think

PR #938, opened 2023-06-15
Allow `mode` in generated Prisma scalar filter types

### LINKED ISSUE #937: [plugin-prisma-utils] Allow `mode: 'insensitive'` for scalar filter types
Prisma allows [case-insensitive filtering](https://www.prisma.io/docs/concepts/components/prisma-client/case-sensitivity#options-for-case-insensitive-filtering) by using the `mode: 'insensitive'` property on a per-field basis.
This is pretty useful for PostgreSQL and MongoDB databases, where case-insensitivity is not enabled by default.

Currently there is no way to add the `mode` field to the generated filter types in Pothos. The [`SchemaBuilder#prismaFilter`](https://github.com/hayes/pothos/blob/f1faf84c051481ea002a47804a1ff466db2890dd/packages/plugin-prisma-utils/src/schema-builder.ts#L42) method does not have an option to enable this field.

Ideally this could be configured just like the other fields in a scalar filter type, e.g.

```ts
const StringFilter = builder.prismaFilter("String", {
  name: "StringFilter",
  ops: ["equals", "not", "contains", "startsWith", "endsWith", "mode"],
});
```

fabsrc:
With this change, the `mode` field can be added to scalar filter types, which allows to do queries like this:

```graphql
{
  users(filter: {email: {contains: "GMAIL", mode: Insensitive}}) {
    id
    email
  }
}
```

I tested the query with the "codegen" example:

```sh
node -r @swc-node/register packages/plugin-prisma-utils/tests/examples/codegen/server.ts
```

Fixes #937

### DIFF
--- a/packages/plugin-prisma-utils/src/schema-builder.ts
+++ b/packages/plugin-prisma-utils/src/schema-builder.ts
@@ -39,7 +39,7 @@ const OrderByRefMap = new WeakMap<
   EnumRef<'asc' | 'desc'>
 >();
 
-const PrismaQueryModeRefMap = new WeakMap<
+const PrismaStringFilterModeRefMap = new WeakMap<
   PothosSchemaTypes.SchemaBuilder<SchemaTypes>,
   EnumRef<'default' | 'insensitive'>
 >();
@@ -87,7 +87,7 @@ schemaBuilder.prismaFilter = function prismaFilter<
             fieldType = [type];
             break;
           case 'mode':
-            fieldType = this.prismaQueryModeEnum();
+            fieldType = this.prismaStringFilterModeEnum();
             break;
           default:
             fieldType = type;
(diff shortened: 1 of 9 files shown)


packages/plugin-prisma-utils/src/schema-builder.ts
```diff
@@ -93,6 +107,23 @@ schemaBuilder.prismaFilter = function prismaFilter<
   return ref as never;
 };
 
+schemaBuilder.prismaQueryModeEnum = function prismaQueryModeEnum() {
```
hayes, line 110: can this be called something less generic, maybe `prismaStringFilterMode` or something, open to suggestions

hayes:
This looks great, thanks for adding this!

packages/plugin-prisma-utils/src/schema-builder.ts
fabsrc: Sure, I adopted the enum name from the Prisma [query-engine](https://github.com/prisma/prisma-engines/blob/489dea78e5d64611960acfde15ac92fee3f21c41/query-engine/connectors/query-connector/src/filter/scalar/mod.rs#L30).

As the enum is defined in the `ScalarFilter` class there, maybe `prismaScalarFilterQueryModeEnum` would be a fitting name here.

hayes: Naming is hard.  I think externally, I would want to avoid having Prisma in the typename (prisma is an implementation detail a consumer of the API shouldn't care about).    I'm a little surprised prisma uses such a generic name, but just `QueryMode` as the type name is likely too generic to use here, I think `StringFilterMode` (mode on `StringFilter`) might be good for the public name.  Maybe `prismaStringFilterModeEnum` as the method name

hayes: I think scalar is probably not the right term here, since this only applies to strings

hayes (approved):

Issue #723, opened 2022-12-27
Support Client Directives in Prisma plugin


hayes:
The prisma plugin should support the ~`@skip`! ~`@include`! and `@defer` directives.

For skip and include, we can return early to exclude the fields selection if the field will not be resolved here: https://github.com/graphql/graphql-js/blob/main/src/execution/collectFields.ts#L250

For defer, it's slightly more complex because we only want to exclude it during the initial execution phase.  I am not yet sure how to determine if a resolver is running deferred or not.

`@stream` may be useful as well for list or connection fields, but would likely be more complicated to implement

juliobetta:
I came across this issue today... I think it would be nice to support at least `@skip` and `@include` first, which are the most common used directives. would you mind sharing the place where that early return would be? thanks!

hayes:
@juliobetta likely it would need to be added somewhere in this function that maps a query to a prisma selection: https://github.com/hayes/pothos/blob/main/packages/plugin-prisma/src/util/map-query.ts#L38

You can see the logic the graphql execute flow uses here: https://github.com/graphql/graphql-js/blob/main/src/execution/collectFields.ts#L306

liquiad:
I think this issue should stay open as `@defer` and `@stream` are still missing.

PR #1017, opened 2023-09-07
Support Client Directives in Prisma plugin (@skip and @include)

julioxavierr:
Related to https://github.com/hayes/pothos/issues/723

### DIFF
--- a/packages/plugin-prisma/tests/index.test.ts
+++ b/packages/plugin-prisma/tests/index.test.ts
@@ -113,6 +113,80 @@ describe('prisma', () => {
     `);
   });
 
+  it('includes fields based on @skip and @include directives', async () => {
+    const query = gql`
+      query {
+        me {
+          id
+          profile @skip(if: false) {
+            bio
+          }
+          posts(limit: 1) @include(if: true) {
+            id
+          }
+        }
+      }
+    `;
+
+    const result = await execute({
+      schema,
+      document: query,
+      contextValue: { user: { id: 1 } },
+    });
+
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "data": {
+          "me": {
+            "id": "VXNlcjox",
+            "posts": [
+              {
+                "id": "250",
+              },
+            ],
+            "profile": {
+              "bio": "Debitis perspiciatis unde sunt.",
+            },
+          },
+        },
+      }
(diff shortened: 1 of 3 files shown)


juliobetta (approved):

hayes (approved):
This looks good to me!  thanks for adding this ❤️

hayes:
It would be great to add a test for these directives.  The prisma tests all verify the queries that are executed, so adding a test that skips some relation field should be pretty straight forward

packages/plugin-prisma/tests/index.test.ts
```diff
@@ -65,6 +65,50 @@ describe('prisma', () => {
     `);
   });
 
+  it('skips fields based on @skip and @include directives', async () => {
+    const query = gql`
+      query {
+        me {
+          id
+          name @skip(if: true)
+          email @include(if: false)
```
hayes, line 74: I think you need to add these to a relation instead to see any difference, unless you query a type that is using select mode.  This change only changes the generated query, and not what revolvers get run.  

```graphql
query {
        me {
          id
          profile @skip(if: true) {
            bio
          } 
          posts(limit: 1) @include(if: false) {
            id
          }

Comment (julioxavierr) [reply packages/plugin-prisma/tests/index.test.ts]: Thanks! I just updated the test

File: packages/plugin-prisma/tests/index.test.ts [hunk 2]
```diff
@@ -65,6 +65,54 @@ describe('prisma', () => {
     `);
   });
 
+  it('skips fields based on @skip and @include directives', async () => {
```
hayes, line 68: sorry to keep asking for more changes one at a time, but would also be nice to check the inverted condition, so that if the conditions is reversed the selection includes the expected relations

julioxavierr: no worries! Just pushed one more commit

packages/plugin-prisma/tests/index.test.ts
hayes: awesome, thanks!

packages/plugin-prisma/tests/index.test.ts
hayes: thanks!

Issue #989, opened 2023-08-18
Prisma connection resolves nodes even when querying only totalCount


julioxavierr:
The Prisma plugin enables using a [prismaConnection](https://pothos-graphql.dev/docs/plugins/prisma#prismaconnection) with a `totalCount` function for loading the total count for the connection. This works great and is useful for containing all info for a relation within the same connection.

However, I recently noticed that even when only querying the `totalCount` for the connection, it still executes the resolver that fetches the edges. And I was wondering if it was possible to avoid that and only execute the `totalCount` resolver.

Following, there is an example based on my current code and an example query. Looking forward to get others thoughts and also open to contribute with a PR.

```ts
builder.prismaObjectField("Account", "measures", (t) =>
  t.prismaConnection({
    type: Measure,
    cursor: "id",
    totalCount: async (source, args) => {
      console.log("🚀 Executed totalCount resolver")
      return prismaClient.measures.count();
    },
    resolve: async (query, source, args) => {
      console.log("🚀 Executed connection resolver")
      return prismaClient.measures.findMany({
        orderBy: { date: { sort: "desc", nulls: "last" } },
        ...query,
      });
    },
  }),
);
```

```graphql
query {
   account(id: "some-id") {
      measures {
         totalCount # should only execute this resolver
      }
   }
}
```

hayes:
this is a bit tricky, as there are many potential ways that optimizing this could inadvertently break things.  A simple example would be adding additional fields or auth checks to the connection object.  Similarly if you have a `nodes` field on the context object, or request anything in `pageInfo`.  In either case, you end up calling functions that get passed the connection results.   I think it would be reasonable to special case the situation where `totalCount` is the only thing selected, but if any other field is selected, the resolver will likely be expecting the full connection result object as the parent value.

Definitely open to a PR for this.    This would likely just be implemented by checking for what fields are selected using the `info` argument, and if the sub-selection is only `totalCount` just returning early with an empty object or something.

hayes:
thanks for fixing this!

julioxavierr:
@hayes I just noticed that this introduced a regression for a specific use case

- A connection is queried differently at the same time, using the field directly in the type and using **a fragment that has a union** (does not happen if the fragment does not use a union)

```graphql
query AppQuery {
  user {
    files {
      totalCount
    }
    ...App_node
  }
}

fragment App_node on Node {
  ... on User {
    id
    files {
      edges {
        node {
          name
          size
        }
      }
    }
  }
[...]

PR #1025, opened 2023-09-11
Prevent executing connection resolver when only total count is selected

julioxavierr:
Related to https://github.com/hayes/pothos/issues/989

### DIFF
--- a/packages/plugin-prisma/src/field-builder.ts
+++ b/packages/plugin-prisma/src/field-builder.ts
@@ -1,4 +1,4 @@
-import { GraphQLResolveInfo } from 'graphql';
+import { getNamedType, GraphQLResolveInfo, isInterfaceType, isObjectType, Kind } from 'graphql';
 import {
   FieldKind,
   FieldRef,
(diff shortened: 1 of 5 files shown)


packages/plugin-prisma/src/field-builder.ts
```diff
@@ -145,6 +145,13 @@ fieldBuilderProto.prismaConnection = function prismaConnection<
           withUsageCheck: !!this.builder.options.prisma?.onUnusedQuery,
         });
 
+        const selection = info.fieldNodes[0];
+        const totalCountOnly =
+          selection.selectionSet?.selections.length === 1 &&
+          selection.selectionSet.selections.every(
+            (s) => 'name' in s && s.name.value === 'totalCount',
+          );
```
hayes, line 153: I think this mostly makes sense, but I think rather than checking the field name, checking for an extension like pothosNoParent or something that indicates the field doesn't rely on the parent value, and then set that extension where the total count field is defined.

hayes (commented):
Thanks so much for the PR!  Sorry it's taken me a bit to getting around to reviewing it.


One thing that this doesn't cover is the total count on related connections.  This isn't a blocker for this PR, but wanted to call out that might also need to be addressed before the issue is fully resolved

packages/plugin-prisma/src/field-builder.ts
julioxavierr: @hayes thanks for the feedback, do you mind sharing an example of this in the codebase so I can use it as a reference?

hayes:
Just pushed some changes that should address the remaining cases

PR #1036, opened 2023-09-21
Update validation plugin docs for Zod validation

ben-bosmans:
Add documentation for providing a Zod schema object to validate a whole field. When I was trying to implement this, it was not previously clear that I could, so hopefully this will make it a little bit clearer.


hayes (approved):
thanks!

PR #1039, opened 2023-09-21
Update error plugin documentation to reference correct plugin names

nathanchapman:
Reference correct plugin names


hayes (approved):
thanks!

PR #1040, opened 2023-09-21
Fix plugin names missing as const

nathanchapman:
From https://pothos-graphql.dev/docs/guide/writing-plugins#indexts:
> The plugins name, which should be typed as a string literal rather than as a generic string:
> const pluginName = 'example' as const;


hayes:
I think the docs there is slightly wrong about `as const` being needed.  Typescript is correctly inferring these, but the extra assertion doesn't hurt anything, thanks!

PR #1057, opened 2023-09-26
Support for adding type mappings in SchemaTypes for Input objects

nathanchapman:


hayes (approved):
Thanks for adding this!!

PR #1079, opened 2023-11-02
feat: allow prismaFilters to filter for null values with not

chenzn1:
`isNot` is used only for relation filter conditions.

### DIFF
--- a/packages/plugin-prisma-utils/tests/index.test.ts
+++ b/packages/plugin-prisma-utils/tests/index.test.ts
@@ -24,7 +24,7 @@ describe('prisma utils', () => {
       query {
         posts(
           order: { author: { name: Desc, profile: null } }
-          filter: { id: { not: { equals: 11, not: null } } }
+          filter: { id: { not: { equals: 11 } } }
         ) {
           id
           author {
@@ -86,7 +86,6 @@ describe('prisma utils', () => {
               "id": {
                 "not": {
                   "equals": 11,
-                  "not": undefined,
                 },
               },
             },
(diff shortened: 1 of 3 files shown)


hayes (approved):

chenzn1:
@hayes Can you merge this?

hayes:
It looks like tests are failing and the PR is missing a changeset

hayes:
In the middle of moving, so can't fix this myself tonight, but might have some time tomorrow

chenzn1:
> It looks like tests are failing and the PR is missing a changeset

The test error existed before, i cant try to fix it.

chenzn1:
> In the middle of moving, so can't fix this myself tonight, but might have some time tomorrow

okay,

packages/plugin-prisma-utils/tests/index.test.ts
```diff
@@ -24,7 +24,7 @@ describe('prisma utils', () => {
       query {
         posts(
           order: { author: { name: Desc, profile: null } }
-          filter: { id: { not: { equals: 11, not: null } } }
```
hayes, line 27: Why did this need to be removed?

hayes:
Thanks for adding the changeset!

packages/plugin-prisma-utils/tests/index.test.ts
chenzn1: because the `id` field not be null

chenzn1: image  

Prisma only allow use null filter on null field,  sush as the `content` field

hayes:
thanks for the fix!

PR #1090, opened 2023-12-03
Add arguments support for prisma relation count

s1owjke:
Pothos supports querying for relation counts with conditions, but we cannot use arguments for this field in graphql which is very useful in practice, this PR add this feature

In [4.16.0](https://github.com/prisma/prisma/releases/tag/4.16.0) the ability to filter the relation count became generally available, so we no longer need `filteredRelationCount`, this section updated in documentation

Here is the example of usage

```typescript
import { builder } from "src/graphql/builder";
import { CommentWhere } from "../inputs/errorLog/where";

export const Client = builder.prismaObject('Post', {
  fields: (t) => ({
    // ...
    commentsCount: t.relationCount('comments', {
      args: { where: t.arg({ type: CommentWhere, required: false }) },
      where: (args) => args.where || {},
    }),
  }),
});
```

And this is an example of a graphql query

```graphql
query PostList {
  postList(take: 20) {
    id
    name
    recentComments: commentsCount(where: {createdAt: {gte: "2023-12-02T12:53:38.065Z"}})
  }
}
```


packages/plugin-prisma/tests/__snapshots__/index.test.ts.snap
```diff
@@ -174,615 +174,6 @@ type Profile {
   user: User
 }
 
-type Query {
```
s1owjke, line 177: Even if you run a snapshot update in `main` branch these changes will be present

hayes (approved):

PR #1099, opened 2023-12-11
test: add enum tests across enum creation options

Liam-Tait:
This pull request introduces tests designed to validate that enum types created using different structures are equivalent.

Each test scenario involves creating an enum type, integrating it into the query type of the schema, and then running two tests:
Introspection Test: Ensures that the schema correctly exposes enum values through GraphQL introspection.
Query Test: Validates the functionality of the enums for type and input.


hayes (approved):
Looks great.  Thanks for the tests!

PR #1103, opened 2023-12-12
docs: add docs for using object as const for enumType

### LINKED ISSUE #591: Document how to extract enum values from object keys
It is common to want to create an enum from an object type, would be good to have some docs on this (`Object.keys(object) as keyof typeof object[]`)

Liam-Tait:
Modern TypeScript may prefer using objects with as const over enums to align with JavaScript
standards. (or my personal preference `Object.freeze`).

Added a primary example using `Object.values` and a secondary example with `Object.keys`
with `enumType`

Added tests for both to prove documented method works

Closes #591

### DIFF
--- a/website/pages/docs/guide/enums.mdx
+++ b/website/pages/docs/guide/enums.mdx
@@ -71,8 +71,8 @@ Enums can be defined a number of different ways:
    Using a values object like this enables defining additional options like a description for each
    enum value.
 
-   Using a values object allows the name of the enum value to be different from the typescript value
-   used internally in your resolvers.
+   Using a values object also allows the name of the enum value to be different from the typescript
+   value used internally in your resolvers.
 
    The keys (eg `Southern`) are used as the name of the enum value in you GraphQL schema, and the
    `value` (eg. `'giraffa'`) property is used as the value you will receive in the arguments for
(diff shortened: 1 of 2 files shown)


website/pages/docs/guide/enums.mdx
```diff
@@ -71,6 +71,44 @@ Enums can be defined a number of different ways:
    Using a values object like this enables defining additional options like a description for each
    enum value.
 
+1. Using an object with `as const`
```
hayes, line 74: Maybe something along the lines of use the values of an existing object type, and similarly mentioned using the keys of the of the object instead.

In general the current docs don't clearly explain the differences between enum names and values and why you may or may not want them to be the same.   I think these examples blur that line a bit, and I want to be explicit here that this will be using the key or value as both name and value.


I think the most useful pattern is actually to go from `{ Foo: 'foo' }` -> `{ Foo: { value: 'foo' } }`, because internally you probably want to use `return MyType.foo`, and in the schema see `enum MyType { Foo }`

hayes (approved):
Thanks again for adding more tests and docs!  I think this overall looks great.  Left one comment (more due to my own omission in the original docs).  I can also probably find some time tomorrow to address that

Issue #1125, opened 2024-01-07
Unable to customise federation directives


TimboTambo:
The Federation plugin does not currently allow the imported federation directives to be overridden when calling `toSubGraphSchema`, which is necessary for 2 main use cases:
1. When the project is using a lower federation version than the package. The link URL can be overridden but the composition of the schema will fail if the package references directives that are not in the federation version of the link URL (e.g. the link URL is at version 2.5, but the package imports `requiresScopes`, which is only in version 2.6 and above).
2. When the project is using a higher version than the package default and wants to import new directives that are not included in the package list.

PR #1126, opened 2024-01-07
Allow override of federation directives

TimboTambo:
See [issue 1125](https://github.com/hayes/pothos/issues/1125).

### DIFF
--- a/packages/plugin-federation/src/schema-builder.ts
+++ b/packages/plugin-federation/src/schema-builder.ts
@@ -14,7 +14,7 @@ import { entitiesField, EntityType, serviceField } from '@apollo/subgraph/dist/t
 import SchemaBuilder, { MaybePromise, SchemaTypes } from '@pothos/core';
 import { ExternalEntityRef } from './external-ref';
 import { Selection, SelectionFromShape, selectionShapeKey } from './types';
-import { entityMapping, mergeDirectives } from './util';
+import { entityMapping, getUsedDirectives, mergeDirectives } from './util';
 
 const schemaBuilderProto = SchemaBuilder.prototype as PothosSchemaTypes.SchemaBuilder<SchemaTypes>;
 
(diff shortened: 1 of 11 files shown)


packages/plugin-federation/src/schema-builder.ts
```diff
@@ -64,18 +79,7 @@ schemaBuilderProto.toSubGraphSchema = function toSubGraphSchema({
...
-              '@requires',
-              '@external',
-              '@extends',
-              '@override',
-              '@authenticated',
-              '@policy',
-              '@requiresScopes',
+              ...federationDirectives,
               options.composeDirectives ? '@composeDirective' : null,
               hasInterfaceObjects ? '@interfaceObject' : null,
```
hayes, line 84: I think potentially it might make more sense to have `authenticated`, `requiresScopes`, and `policy`, and maybe even all the directives work like `composeDirective` and `interfaceObject` where they are only imported when used.  I think being able to explicitly add your own directives is good, but I think we should potentially just auto-import any directive that gets used through a plugin feature

hayes (commented):
Thanks for the PR!  I think in general this makes sense, but I think this would be a good time to make things a little more consistent between features (right now `composeDirective` and `interfaceObject` are treated differently without any specific reason

packages/plugin-federation/src/schema-builder.ts
TimboTambo: Sure, makes sense - is there a simple way to derive at this point which of these directives have been used, either via (e.g.) the `requiresScopes` option or the directives array on each query/ mutation/ enum/ etc...?

hayes: Not yet, I can try to find some time in the next few days to add something like that.  I think it would probably be easiest to track that as the directives are added into the schema in the plugin class

hayes (approved):

hayes:
Added some logic to track what directives are used, and defaulted the imported directives to only import what is used

website/pages/docs/plugins/federation.mdx
```diff
@@ -265,7 +264,7 @@ builder.asEntity(Media, {
 
 See federation documentation for more details on `interfaceObject`s
 
-### composeDirective
+### composeDirective =
```
TimboTambo, line 267: Is this `=` a typo?

TimboTambo:
> Added some logic to track what directives are used, and defaulted the imported directives to only import what is used

Awesome! Going to be using this straight away. Thanks.

website/pages/docs/plugins/federation.mdx
hayes: Yup, not sure where that come from

hayes: Looks like it was already in the Readme, which I copied here to keep it in sync

Issue #1141, opened 2024-01-26
[Feature Request] enhance smart subscriptions plugin design to support cloudflare workers


jb747:
The cloudflare workerd design does not allow the SubscriptionManager.pushValue() to respond to a subscription request during a mutation request because the I/O is crossing request boundaries. The cloudflare worker error:

```text
Error: Cannot perform I/O on behalf of a different request. I/O objects (such as streams, request/response bodies, and others) created in the context of one request handler cannot be accessed from a different request's handler. This is a limitation of Cloudflare Workers which allows us to improve overall performance.
```

There is also an issue with setTimeout, I suspect because it does not resolve within the current request context. This can be temporarily hacked-to-work via a polyfill which skips the timeout if the debounceDelay is set to 0.

There may be other issues. Just FYI.

jb747:
I was using Yoga pubsub to detect-and-trigger events. A [less-coupled mechanism](https://github.com/bubblydoo/graphql-workers-subscriptions) seems promising... looking into it.

hayes:
I'm not very familiar with cloudflare workers, but the smart-subscriptions plugin isn't tied to any specific type of IO, so using something like the library you linked above is probably the right path forward.  I don't think any changes should be required in the plugin itself, as it is intended to be decoupled from the actual pub-sub mechanism. 

There is only one setTimeout used in the smart-subscriptions plugin, and it can be bypassed by setting `debounceDelay` to null in the plugins options

jb747:
Just added a minor pull request to allow skipping the timeout if debounceDelay is null. Hopefully it helps..

hayes:
Just merged the PR, thanks!

PR #1143, opened 2024-01-27
if debounceDelay is null, keep it null

jb747:
as per [intended design](https://github.com/hayes/pothos/issues/1141).


hayes (approved):
Nice, thanks for the fix!

hayes:
Just need a changeeset to get this released

PR #1189, opened 2024-04-16
[auth] Add RequestCache.clearForContext and export RequestCache class

awinograd:
If the scope initializer loads data that may change during a request for example due to a new user session or permission change then we need a way to clear the cache such that the scope initializer can be re-executed based on the new context state.

@hayes this contribution is based on our discord discussion: https://discord.com/channels/625400653321076807/868135255616417834/1227683491492135092

To be honest, I didn't fully grasp your message [I'd probably also update...](https://discord.com/channels/625400653321076807/868135255616417834/1227684722218631340) in terms of what you wanted to update. I will note that I ended up adding a new static method instead of a `clearCache()` instance method you were proposing as it seemed easier than to manually reset/reinitialize each of the instances internal caches. That said, I'm happy to update this PR as much as needed to get it to your liking.

### DIFF
--- a/.changeset/clever-numbers-jog.md
+++ b/.changeset/clever-numbers-jog.md
@@ -1,0 +1,5 @@
+---
+'@pothos/plugin-scope-auth': minor
+---
+
+[auth] Allow clearing/resetting scope cache in the middle of a request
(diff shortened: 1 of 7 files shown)


hayes:
Pushed up some changes here https://github.com/awinograd/pothos/pull/1 that adds the change to use createContextCache

hayes (approved):
Thanks for adding this!

awinograd:
Thank you for the guidance @hayes !

PR #1201, opened 2024-05-19
add support for progressive override in federation

RomKadria:
hey, we use your lib and need support for the progressive override feature as seen here 
https://www.apollographql.com/docs/federation/entities-advanced/#incremental-migration-with-override

Thanks :)

### DIFF
--- a/.changeset/small-ravens-guess.md
+++ b/.changeset/small-ravens-guess.md
@@ -1,0 +1,5 @@
+---
+"@pothos/plugin-federation": minor
+---
+
+Add label option to the @override directive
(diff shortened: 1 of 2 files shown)


hayes (commented):
I think label needs to be optional, but otherwise this looks good! Can you also add a changesets?

hayes (approved):
Thanks!

hayes:
Sorry, i probably should have been a little more specific.  This repo uses changesets to manage releases/versioning (instructions in the comment from the changesets bot above),  so we need to revert the change log entry and the version change, but add a changesets entry.  I can also do this myself later when I'm back at a computer

RomKadria:
> Sorry, i probably should have been a little more specific. This repo uses changesets to manage releases/versioning (instructions in the comment from the changesets bot above), so we need to revert the change log entry and the version change, but add a changesets entry. I can also do this myself later when I'm back at a computer

Hey sorry didnt see it was away for a bit. Thanks so much!!!

PR #1214, opened 2024-06-15
WIP: Improve dataloader type infering

### LINKED ISSUE #951: Dataloader infers wrong type from load method when returning non-base errors
> The typing for the load method is not ideal, it basically is typed as `load: (some args) => MaybePromse<Error | Shape>` where `Shape` a generic without any constraints. This means if load returns `Promise<User | Error>` `Shape` is inferred as being `User.`  The better way to handle this would be to have load be defined as something like `load: (some args) => MaybePromse<LoadResult>` and then other places that use Shape would be updated to do something like `ShapeFromLoadResult<LoadResult>` instead of Shape.  `ShapeFromLoadResult` would probably be something like type `ShapeFromLoadResult<T> = T extends Error ? never : T`

iamchanii:
Hello! I'm happy to there's something looks like what I can. If you still have an interest about #951, give me some feedbacks! of course, you can feel free close this if already fixed or planned in days!

## Description

- Improve `load` method type inference from dataloader plugin.
- Resolve #951

## What I did

- Add `ShapeFromLoadResult<T>` to inferring only result type.
- Add `LoadResult` type parameter to loadable* methods from SchemaBuilder.
- Change `Shape` type parameter's constraint.

## To Do

- [ ] Update codes in `deno`

### DIFF
--- a/.changeset/rare-mayflies-yell.md
+++ b/.changeset/rare-mayflies-yell.md
@@ -1,7 +1,0 @@
----
-"@pothos/plugin-tracing": minor
-"@pothos/tracing-sentry": minor
-"@pothos/deno": minor
----
-
-Update sentry plugin to be compatible with sentry v8
--- a/examples/open-telemetry/CHANGELOG.md
+++ b/examples/open-telemetry/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @pothos-examples/open-telemetry
 
+## 1.2.42
+
+### Patch Changes
+
+- Updated dependencies [cea5c66]
+  - @pothos/plugin-tracing@0.6.0
+  - @pothos/tracing-opentelemetry@0.6.9
+
 ## 1.2.41
 
 ### Patch Changes
--- a/examples/open-telemetry/package.json
+++ b/examples/open-telemetry/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pothos-examples/open-telemetry",
-  "version": "1.2.41",
+  "version": "1.2.42",
   "main": "index.js",
   "license": "MIT",
   "dependencies": {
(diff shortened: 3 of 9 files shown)


packages/plugin-dataloader/src/field-builder.ts
```diff
@@ -187,7 +193,7 @@ fieldBuilderProto.loadableGroup = function loadableGroup<
       const values = await load(keys, ctx, args as never);
       const groups = new Map<Key, ShapeFromTypeParam<SchemaTypes, Type, true>[]>();
 
-      for (const value of values) {
+      for (const value of values as ShapeFromTypeParam<SchemaTypes, Type, true>[]) {
```
iamchanii, line 196: I don't know how to type it better than this

packages/plugin-dataloader/src/types.ts
```diff
@@ -43,7 +45,7 @@ export type LoadableFieldOptions<
     keys: Key[],
     context: Types['Context'],
     args: false extends ByPath ? never : InputShapeFromFields<Args>,
-  ) => Promise<readonly (Error | LoaderShapeFromType<Types, Type, Nullable>)[]>;
+  ) => MaybePromise<LoadResult>;
```
iamchanii, line 48: did I used `MaybePromise<T>` correctly here?

hayes:
Thank you for opening this PR! Sorry for not getting you a review sooner.  I will try to review the or this weekend

hayes:
Sorry for closing this out, feel free to pull in the changes from #1221 and re-open this.  I was trying to push some changes to your PR, and accidentally mixed up the git commands.  #1221 updated some of your changes to re-order some of the generic arguments, and simplify things a bit, but I think overall, this seems pretty close to what we wanted!

Thanks again for the contribution!

PR #1216, opened 2024-06-21
Docs: Make details panel non-interactive when collapsed

lubosmato:
Collapsed details panel covered content due to full height. 

The change makes details panel to be full height only when is open and thus improves UX of copying code snippets that would have been covered by the details panel (see gif below).

code-select


hayes:
Thanks!

PR #1230, opened 2024-07-09
Add PrismaTypesFromClient utility type

2coo:
Solves: https://github.com/hayes/pothos/discussions/1193

[`PrismaTypesFromClient`](https://gist.github.com/2coo/11df1d175444e922571fab8c7e1ce6af) utility type that infers types from `PrismaClient`.

Example:
```typescript
import SchemaBuilder from "@pothos/core"
import type { PrismaTypesFromClient } from "@pothos/prisma-plugin"
import { BigNumberExtension } from "~/lib/prisma-extensions"

const db = new PrismaClient().$extend(BigNumberExtension)

export const builder = new SchemaBuilder<{
  PrismaTypes: PrismaTypesFromClient<typeof db>
}>()
```


hayes:
This great, thanks for adding this, I'll get it out as part of 4.0!

PR #1249, opened 2024-07-17
docs: update documentation about onFirstUse method

iamchanii:
First of all, I'd like to congrats on the release of v4!

While migrates some codes, I found `builer.configStore.onFieldUse` has been removed, and instead the `onFirstUse` method on `FieldRef` has been added. so I made a little change in Writing Plugins document.


hayes (approved):
Thanks!

PR #1266, opened 2024-08-02
Remove @sentry/tracing from @pothos/sentry-tracing

penpendayo:
The dependency on @sentry/tracing has been removed in this PR:
https://github.com/hayes/pothos/pull/1220/files#diff-ebc36762ab958f662687f4e52e77f8e99c4765d7bffea474d5936f2890f8e893L53

It seems it was mistakenly added during the v4 release:
https://github.com/hayes/pothos/pull/855/files#diff-ebc36762ab958f662687f4e52e77f8e99c4765d7bffea474d5936f2890f8e893R53

Since Sentry v8, @sentry/tracing has been removed, so it is correct to not have this dependency.

### DIFF
--- a/.changeset/swift-yaks-shake.md
+++ b/.changeset/swift-yaks-shake.md
@@ -1,0 +1,5 @@
+---
+"@pothos/tracing-sentry": patch
+---
+
+Remove @sentry/tracing from @pothos/sentry-tracing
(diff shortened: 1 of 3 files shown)


hayes:
Sorry about that, thanks for the fix!

PR #1272, opened 2024-08-06
add missing swc option

JipSterk:
hi @hayes, 

As said this week i would checkout the newly created drizzle plugin for pothos. 
However when i installed it next threw me an error that the module was missing. 
The following change should fix the output not being `lib/src/...` but just `lib/...` as described in the [package.json](https://github.com/hayes/pothos/blob/main/packages/plugin-drizzle/package.json)


hayes (approved):
thanks!

PR #1292, opened 2024-09-07
chore: export connection utilities + types

lkuoch:
This will be extremely helpful as I am writing relay connections for one of my projects and want to reuse this method to calculate relay cursor arguments from default connection args.

### DIFF
--- a/packages/plugin-relay/package.json
+++ b/packages/plugin-relay/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pothos/plugin-relay",
-  "version": "4.2.0",
+  "version": "4.2.1",
   "description": "A Pothos plugin for adding relay style connections, nodes, and cursor based pagination to your GraphQL schema",
   "main": "./lib/index.js",
   "types": "./dts/index.d.ts",
--- a/packages/plugin-relay/package.json
+++ b/packages/plugin-relay/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pothos/plugin-relay",
-  "version": "4.2.1",
+  "version": "4.2.0",
   "description": "A Pothos plugin for adding relay style connections, nodes, and cursor based pagination to your GraphQL schema",
   "main": "./lib/index.js",
   "types": "./dts/index.d.ts",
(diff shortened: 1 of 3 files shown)


packages/plugin-relay/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pothos/plugin-relay",
-  "version": "4.2.0",
+  "version": "4.2.1",
```
hayes, line 3: Versioning is handled automatically through the changeset release flow

packages/plugin-relay/src/utils/connections.ts
```diff
@@ -202,7 +202,7 @@ export function resolveArrayConnection<T>(
   };
 }
 
-function parseCurserArgs(options: ResolveOffsetConnectionOptions) {
+export function parseCurserArgs(options: ResolveOffsetConnectionOptions) {
```
hayes, line 205: This should probably be renamed to parseCursorConnectionArgs before being exported

packages/plugin-relay/src/utils/connections.ts
```diff
@@ -231,7 +231,7 @@ function parseCurserArgs(options: ResolveOffsetConnectionOptions) {
 }
 
 // biome-ignore lint/suspicious/noRedeclare: <explanation>
-type NodeType<T> = T extends (infer N)[] | Promise<(infer N)[] | null> | null ? N : never;
+export type NodeType<T> = T extends (infer N)[] | Promise<(infer N)[] | null> | null ? N : never;
```
hayes, line 234: This would need a much less generic name to be exported, and I'm not sure if it actually makes sense to export as is.

packages/plugin-relay/src/utils/connections.ts
```diff
@@ -202,7 +202,7 @@ export function resolveArrayConnection<T>(
   };
 }
 
-function parseCurserArgs(options: ResolveOffsetConnectionOptions) {
+export function parseCursorCurserArgs(options: ResolveOffsetConnectionOptions) {
```
hayes, line 205:
```suggestion
export function parseCursorConnectionArgs(options: ResolveOffsetConnectionOptions) {
```

hayes (approved):

PR #1320, opened 2024-10-21
Fix ref relay options in plugin-prisma

maruware:
This PR is a fix that corresponds to the change of relay option in v4

### DIFF
--- a/.changeset/cold-cups-love.md
+++ b/.changeset/cold-cups-love.md
@@ -1,0 +1,5 @@
+---
+"@pothos/plugin-prisma": patch
+---
+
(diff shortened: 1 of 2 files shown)


hayes (approved):
Thanks for the fix!  I recently added a builder.nodeRef method to the relay plugin that can be used to avoid having to re-implement node objects in other plugins, so I updated your PR to use that instead

tmaruyama-bitkey:
Thank you! I'm glad it's been improved.

maruware:
Thank you! I'm glad it's been improved.

PR #1343, opened 2024-11-12
fix: drizzle plugin model loader when multiple models are requested

olup:


packages/plugin-drizzle/src/model-loader.ts
```diff
@@ -178,7 +178,7 @@ export class ModelLoader {
           ...selectionToQuery(selection),
           where: inArray(
             this.selectSQL,
-            [entry.models.keys()].map(([model]) => this.sqlForModel(model)),
+            [...entry.models.keys()].map((model) => this.sqlForModel(model)),
```
hayes, line 181: good find!

hayes (approved):
thanks for fixing this!

PR #1348, opened 2024-11-15
upgrade to  v17

yaacovCR:
with canary-pr-4288

### DIFF
--- a/examples/complex-app/package.json
+++ b/examples/complex-app/package.json
@@ -27,13 +27,16 @@
     "@pothos/plugin-simple-objects": "workspace:*",
     "@pothos/plugin-zod": "workspace:*",
     "@prisma/client": "^5.19.1",
-    "graphql": "^16.8.1",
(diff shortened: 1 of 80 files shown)


yaacovCR:
proof of concept, see: https://github.com/graphql/graphql-js/pull/4288#issuecomment-2478343428

besides the caveats mentioned there, this version of the upgrade does not preserve backwards compatibility

not sure you will find this helpful, but putting it out there (rather than just deleting after my experimentation) just in case

side-note: this project is amazing!

hayes:
thanks for sharing this! This will be really useful

I think this will end up needing to be a major release to handle the defaultValue changes properly

yaacovCR:
> I think this will end up needing to be a major release to handle the defaultValue changes properly

This comment inspired https://github.com/graphql/graphql-js/pull/4296 -- not 100% sure, but I wonder if that means that the upgrade to v17 no longer has to be breaking (at least for this library).

PR #1351, opened 2024-11-25
Bind auth scope loaders to scopes object

rrdelaney:
This PR binds scope loaders to the `scopes` object itself before execution. The motivation here is to allow passing in external classes that run authorization checks, such as the following pattern we tried when configuring `@pothos/plugin-scope-auth`:

```tsx
export const builder = new SchemaBuilder<{
  Context: ResolverContext;
  AuthScopes: {
    canEditObject: string;
  };
}>({
  plugins: [ScopeAuthPlugin],
  scopeAuth: { authScopes: (ctx) => ctx.auth },
});
```

where `ctx.auth` is an instance of a class that performs authorization checks e.g: 

```tsx
class AuthorizationChecker {
  private readonly user: User;
  canEditObject(objectUuid: string): Promise<boolean> {
    // Performs some checks based on objectUuid and this.user.
  }
}
```

Although this only saves us a few lines of code, it prevents issues where the return type of `authScopes` matches what the typechecker expects, but fails at runtime due to binding issues.

### DIFF
--- a/.changeset/nervous-shoes-grin.md
+++ b/.changeset/nervous-shoes-grin.md
@@ -1,0 +1,5 @@
+---
+"@pothos/plugin-scope-auth": patch
+---
+
+Bind `authScopes` loaders to returned provider
--- a/packages/plugin-scope-auth/src/request-cache.ts
+++ b/packages/plugin-scope-auth/src/request-cache.ts
@@ -159,7 +159,7 @@ export default class RequestCache<Types extends SchemaTypes> {
     const key = this.cacheKey ? this.cacheKey(arg) : arg;
 
     if (!cache.has(key)) {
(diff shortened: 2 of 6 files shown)


hayes:
fwiw, you can change you class slightly so that its methods are bound: ` canEditObject: (objectUuid: string): Promise<boolean> => {`

hayes (approved):

rrdelaney:
> This seems reasonable, thanks for adding tests!
> 
> Just need a changeset to be able to merge/release this

Thanks! Added a changeset for this, marked it as a patch version.

And yep we just used the arrow function workaround for now, figured I'd upstream a change to prevent surprising runtime behavior when typechecks pass for new users too

Issue #1370, opened 2025-01-05
drizzleTableKey symbol reference not exported by `@pothos/plugin-drizzle`


m1212e:
### Discussed in https://github.com/hayes/pothos/discussions/1368

The drizzle plugin is not exporting a reference to the `drizzleTableKey` symbol used in https://github.com/hayes/pothos/blob/9dfc0c198b422bb8726d86dec20df339e071a508/packages/plugin-drizzle/src/object-ref.ts#L12

In my scenario I'm building a library around this dependency. When I use the `ObjectRef` type TS expects a field with a key which is the symbol of https://github.com/hayes/pothos/blob/9dfc0c198b422bb8726d86dec20df339e071a508/packages/plugin-drizzle/src/object-ref.ts#L10-L11 which isn't possible since the type is not exported by `@pothos/plugin-drizzle`.

This results in TS redeclaring the type like this upon processing the .ts source files:
```ts
declare const drizzleTableKey: unique symbol;
declare class DrizzleObjectRef<Types extends SchemaTypes, Table extends keyof Types["DrizzleRelationSchema"] = keyof Types["DrizzleRelationSchema"], T = {}> extends ObjectRef<Types, T> {
    [drizzleTableKey]: Types["DrizzleRelationSchema"][Table];

    ...
}
```
which obviously is a different type reference than whats used by the library internally in function declarations like `t.drizzleField` and results in a type error.

In my case this means that I cannot export a function in my library that returns a `DrizzleObjectRef` of the correct type because TS redeclares that type when I build my library because it cannot use the non exported type of `@pothos/plugin-drizzle`. An export of that type would solve this issue.

hayes:
PR looks good, I'll get that released for you!

hayes:
just kicked off the release, should be out in a few minutes.  Thanks for the fix!

m1212e:
Thank you very much!

PR #1371, opened 2025-01-05
Export Object ref type in drizzle plugin

m1212e:
Addresses https://github.com/hayes/pothos/issues/1370 by exporting the type in the `types` module


hayes (approved):

hayes:
thanks, this looks good to me!

PR #1373, opened 2025-01-07
add new validation plugin

hayes:


packages/core/src/index.ts
```diff
@@ -203,12 +210,37 @@ export const ListRef = InternalListRef as new <Types extends SchemaTypes, T, P =
...
+  initConfig: (name: string, typeConfig: PothosTypeConfig) => PothosOutputFieldConfig<Types>,
+) => PothosSchemaTypes.FieldRef<Types, T, Kind>;
+
+export type InputFieldRef<Types extends SchemaTypes, T> = PothosSchemaTypes.InputFieldRef<Types, T>;
+export const InputFieldRef = InternalInputFieldRef as new <Types extends SchemaTypes, T>(
+  initConfig: (name: string, typeConfig: PothosTypeConfig) => PothosInputFieldConfig<Types>,
+) => PothosSchemaTypes.InputFieldRef<Types, T>;
+
+export type ArgumentRef<Types extends SchemaTypes, T> = PothosSchemaTypes.ArgumentRef<Types, T>;
+export const ArgumentRef = InternalArgumentRef as new <Types extends SchemaTypes, T>(
```
hayes, line 233: This just makes it so these types of Refs can be more easily extended by plugins

hayes:
Just pushed up an implementation of the main validation logic.  I think this plugin is now mostly functional, but needs testing, docs, and some config for error handling

packages/plugin-validation/src/types.ts
```diff
@@ -0,0 +1,14 @@
+import type { SchemaTypes } from '@pothos/core';
+import type { GraphQLResolveInfo } from 'graphql';
+import type { ZodError } from 'zod';
```
hayes, line 3: should not use zod

packages/plugin-validation/src/utils.ts
```diff
@@ -0,0 +1,229 @@
...
+            issues,
+          }
+        : {
+            value: mapped,
+            issues: undefined,
+          };
+    });
+  };
+}
+function reduceMaybeAsync<T, R>(
```
hayes, line 213: duplicated from core

PR #1383, opened 2025-02-26
fix(add-graphql): pass through isOneOf setting from input objects

acburdine:
this ensures native GraphQLInputObjectTypes with isOneOf enabled are correctly added to the pothos schema


hayes (approved):

hayes:
thanks for the fix!

acburdine:
happy to help! Pothos has been great to work with 😄

PR #1405, opened 2025-03-27
[prisma, drizzle] skip selections on `@defer`

XiNiHa:
This PR makes the Prisma and Drizzle plugins skip fetching ahead if `@defer`red fragment is found, which matches the expected behavior of `@defer`.
It's also configurable with the `skipDeferredFragments` option of each plugin.


packages/plugin-drizzle/src/utils/map-query.ts
```diff
@@ -199,15 +220,29 @@ function addNestedSelections(
   state: SelectionState,
   selections: SelectionSetNode,
   indirectPath: string[],
+  skipDeferredFragments: boolean,
```
XiNiHa, line 223: I'm not sure it's a good idea to drill it down here just for this 😵‍💫

PR #1411, opened 2025-04-01
Setup pkg.pr.new

XiNiHa:
This PR configures continuous releases using [pkg.pr.new](https://github.com/stackblitz-labs/pkg.pr.new).

To function properly, [the GitHub application](https://github.com/apps/pkg-pr-new) should be installed on the repo.

### DIFF
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -8,22 +8,6 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    services:
-      # Label used to access the service container
-      postgres:
-        # Docker Hub image
-        image: mghayes/multi-postgres:v2
-        ports:
-          - '5455:5432'
-        env:
-          POSTGRES_USER: prisma
-          POSTGRES_PASSWORD: prisma
-          POSTGRES_MULTIPLE_DATABASES: tests,drizzle
-
-        # Set health checks to wait until postgres has started
-        options: >-
-          --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
-
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,67 @@
...
+            //registry.npmjs.org/:_authToken=$NPM_TOKEN
+          EOF
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Build
+        run: pnpm run ci
```
hayes, line 65: This will run tests and linting.  Do you want publishing to be dependent on that? I don't have a specific opinion on this.  I know for my own prs tests are often a blocker for getting things merged, so if the goal is to test things out before it's merged this might not be what you want

package.json
```diff
@@ -76,7 +77,7 @@
       }
     },
     "patchedDependencies": {
-      "@changesets/assemble-release-plan": "patches/@changesets__assemble-release-plan.patch"
+      "@changesets/assemble-release-plan@6.0.5": "patches/@changesets__assemble-release-plan.patch"
```
hayes, line 80: What's the reason behind this change, is pkg-pr-new also using this (at a different version)

.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,67 @@
...
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    services:
+      # Label used to access the service container
+      postgres:
```
hayes, line 13: This is only needed for testing, so it can be removed if you end up not running tests below

.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,67 @@
...
+        with:
+          node-version: 22
+          cache: 'pnpm'
+          cache-dependency-path: pnpm-lock.yaml
+
+      - name: Setup turbo cache
+        uses: actions/cache@v3
+        with:
+          path: node_modules/.cache
+          key: turbo-${{ runner.os }}-18-${{ hashFiles('**/pnpm-lock.yaml') }}
```
hayes, line 50: I think this 18 used to refer to the node version, which is now out of sync.  

Should this share a turbo cache with other jobs? You might want a different key here

.github/workflows/pkg-pr-new.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Creating .npmrc
+        run: |
+          cat << EOF > "$HOME/.npmrc"
+            email=michael@hayes.io
+            //registry.npmjs.org/:_authToken=$NPM_TOKEN
+          EOF
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
hayes, line 59: You don't need this right?

.github/workflows/pkg-pr-new.yml
XiNiHa: I've intentionally included the tests to be run but now I think it makes more sense to exclude it 😅

package.json
XiNiHa: Not sure, the patch broke on my local so I recreated the patch. If I exclude the version specifier it still doesn't work after the repatch.

.github/workflows/pkg-pr-new.yml
XiNiHa: I thenk so 👍

hayes (approved):

PR #1428, opened 2025-04-25
Drizzle RQBv2 fixes

XiNiHa:


hayes:
Thanks for fixing these!  I am trying to find some time soon to get the drizzle plugin changes merged.  It needs a lot of testing, so it's been hard to find a big enough chunk of time to really focus on it, but should be out soonish

hayes:
I cherry-picked your commits into the branch, because of some very annoying rebases.  Thanks for the fixes!

PR #1440, opened 2025-05-03
Allow entry connections with drizzleConnectionHelper

XiNiHa:
This PR enables using `drizzleConnectionHelper` to create advanced Drizzle connections outside Drizzle objects.


hayes (approved):
Thanks for adding this!

PR #1461, opened 2025-06-15
add cost and list size directives

RomKadria:
Add support to the cost and listsize directive
reference
https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/reference/directives#cost

### DIFF
--- a/.changeset/modern-facts-trade.md
+++ b/.changeset/modern-facts-trade.md
@@ -1,0 +1,5 @@
+---
+"@pothos/plugin-federation": minor
+---
+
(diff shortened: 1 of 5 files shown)


hayes (commented):
Thanks for adding this, I think this looks generally correct!

I think you reformatted the files you edited with a different formatter, can you run biome to fix the styling?


It would also be good to add a basic test.  It doesn't have to be anything too complex, just add some usage of the new directives into the test schema and update the schema snapshots to ensure they are being added correctly

hayes (approved):
looks good, thanks!

PR #1464, opened 2025-06-18
add itemErrors option for erros plugin

hayes:


packages/plugin-prisma/src/generator.ts
```diff
@@ -42,22 +42,22 @@ generatorHandler({
...
-        : prismaOutputLocation.startsWith('./')
-          ? `./${posix.join(prismaOutputLocation, 'index.js')}`
-          : posix.join(prismaOutputLocation, 'index.js');
+    // const prismaLocation =
+    //   prismaOutputLocation.startsWith('@') ||
+    //   extensions.some((ext) => prismaOutputLocation.endsWith(ext))
+    //     ? prismaOutputLocation
+    //     : prismaOutputLocation.startsWith('./')
+    //       ? `./${posix.join(prismaOutputLocation, 'index.js')}`
+    //       : posix.join(prismaOutputLocation, 'index.js');
```
hayes, line 60: I think this is the fix for an issue in the latest prisma release.  It looks like prisma always generates a package.json file with an exports field defined.  I previously was importing from /prisma-path/index.ts instead of just prisma-path because I was worried about esm resolution, but with the package.json exports, I don't think thats actually an issue anymore

packages/plugin-errors/src/index.ts
```diff
@@ -79,38 +86,216 @@ export class PothosErrorsPlugin<Types extends SchemaTypes> extends BasePlugin<Ty
...
+    const { name: getUnionName = defaultUnionName, ...defaultUnionOptions } =
+      builderUnionOptions ?? {};
+
     const {
       types = [],
       result: {
         name: resultName = getResultName({
           parentTypeName,
-          fieldName: capitalize(fieldConfig.name),
+          fieldName,
```
hayes, line 290: this was a bug, but probably needs a changeset since it has the potential to change type names in the schema if people were depending on this capitalization i custom name functions

PR #1510, opened 2025-08-27
Update zod to v4

acburdine:
Updates the zod validation plugin to use zod v4. I originally tried the approach to allow passing in both zod v3 and v4 schemas, but I don't know that it would really work with the dynamic zod schema builder 😕


acburdine:
originally was using the `zod/v4` import paths, but it looks like something in the build process messed with the generated imports which caused the errors package tests to break 🤔

acburdine:
> Are there specific things you would miss from the zod plugin if you migrated to the validation plugin?

Ah, hadn't seen that plugin yet 😅 That plugin works with my usage as far as I can tell - will do a bit more testing today.

acburdine:
The validation plugin works for what I need, so this PR's not needed on my end anymore 👍🏻 that said, happy to finish up this PR (if anything else needs done) or just close it if there's not a desire to update the zod plugin to v4.

or maybe, the best approach would be to update the zod plugin docs to indicate that for zod 4 support it's best to use the new validation plugin? could make a PR to that effect as well 😄

hayes:
The diff here seems smaller than I would have expected! I thought this would be more complicated to make work with v4

I think the responsible thing to do is to make it easy to keep dependencies up to date.  Not supporting zod4 means people need to choose between doing the bigger migration to the validation plugin, or letting their zod dep be out of date, so I am leaning towards letting people update to zod4 instead. I think people updating pothos and its plugins are likely to also want to update zod, so that seems like the best path forward.

I have always kept pothos plugins and core on the same major version, so while this is technically a breaking change, I think the thing to do here would be to update the changeset to be a minor with a big warning, and a recommendation to move to the new validation plugin.  Having plugin-zod at v5 with everything on v4 is probably not worth it, and this change by itself doesn't warrant bumping the whole ecosystem to v5.  I am probably going need a v5 soon for some internal type changes that will require a few 3rd party plugins to update their types, but thats probably a few months out at the earliest

Adding a callout about the new validation plugin to the zod docs page would also be appreciated.

If you want to make those changes I am happy to merge the PR, if not I can probably also take it from here when I have a little extra time

hayes (approved):

PR #1533, opened 2025-09-26
fix graphql playground url

adriablancafort:


hayes (commented):
Did something about this change recently, as far as I know this was is correct

PR #1585, opened 2025-12-24
Allow interface merging for `UserSchemaTypes.Scalars`

### LINKED ISSUE #1584: `UserSchemaTypes.Scalars` isn't extendable from plugins.
I'm writing a custom plugin, which requires the `DateTime` scalar to be installed.

So I added the constraint to the `global-types.ts`

```ts
declare global {
  export namespace PothosSchemaTypes {
    export interface UserSchemaTypes {
      Scalars: {
        DateTime: { Input: unknown, Output: unknown }
      }
    }
  }
}
```

But got this error from tsc

```
error TS2717: Subsequent property declarations must have the same type.  Property 'Scalars' must be of type 'Record<string, { Input
: unknown; Output: unknown; }>', but here has type '{ DateTime: { Input: unknown; Output: unknown; }; }'
```

I guess the `Scalars` should be `{ [key: string]: { Input: unknown; Output: unknown; }` rather than `Record` to allow interface merging.

cometkim:
Fixes #1584

### DIFF
--- a/.changeset/tender-trams-chew.md
+++ b/.changeset/tender-trams-chew.md
@@ -1,0 +1,5 @@
+---
+"@pothos/core": patch
+---
+
+Allow declaration merging for `UserSchemaTypes.Scalars` from plugins.


hayes:
Seems reasonable to me, can you add a changeset?

PR #1587, opened 2025-12-29
add add-graphql option includeSchemaDirecties

cometkim:
While I'm doing incremental migration to Pothos using the add-graphql plugin,

```ts
import SchemaBuilder from '@pothos/core';
import AddGraphQLPlugin from '@pothos/plugin-add-graphql';

import { filterSchema } from '@graphql-tools/utils';
import { schema as orignalSchema } from '#schema.ts';

const remainingSchema = filterSchema(originalSchema, { ... });

const builder = new SchemaBuilder({
  plugins: [AddGraphQLPlugin],
  add: {
    schema: remainingSchema,
  },
});
```

I noticed that directives from `originalSchema` are omitted. Even with `printSchemaWithDirectives()` from `@graphql-tools/utils`, it doesn't generate directives in the SDL.

This was important to me because I needed to verify that the new schema provided the same functionality (including the printer) as the existing schema.

I made this patch to fix it, but I'm not sure it can be merged upstream.

### DIFF
--- a/packages/plugin-add-graphql/package.json
+++ b/packages/plugin-add-graphql/package.json
@@ -48,6 +48,9 @@
     "@pothos/core": "*",
     "graphql": "^16.10.0"
   },
+  "dependencies": {
+    "@graphql-tools/utils": "^10.11.0"
+  },
   "devDependencies": {
     "@pothos/core": "workspace:*",
     "@pothos/test-utils": "workspace:*",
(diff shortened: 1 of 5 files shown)


cometkim:
> What format are the directives in in your current schema that they don't show up in extensions with the current implementation?

`originalSchema` is parsed SDL by `@graphql-tools/schema` `makeExecutableSchema()`

PR #1594, opened 2026-01-14
Support a non-filtered existence field for Prisma related connections

jgnieuwhof:
## Motivation
We'd like to add a 'non-filtered existence check' for Prisma related connections, for two reasons:
- it can be useful to have both a filtered total count, and a non-filtered existence check for UI states
- existence is faster than aggregating a total count for large relations

One example is a searchable table with pagination, with a prompt for the user to create their first record if none exist yet. If a user enters a search term with no matching results it's impossible to know whether to display the 'create first record' prompt _or_ an empty table. A non-filtered `totalCount` would work, however we need the filtered counts for # of results / pagination and the `filterConnectionTotalCount` option is global.

## Approach
I've added a `hasAny` global connection field to our builder as follows:
```typescript
builder.globalConnectionField("hasAny", (t) =>
  t.boolean({
    nullable: false,
    description: "A fast, non-filtered existence check.",
    args: {},
    resolve: async (connection, args, ctx, info) => {
      const parentType = info.schema.getType(
        info.path.prev!.typename!
      ) as GraphQLObjectType;
      const connectionField = parentType.getFields()[info.path.prev!.key];
      const prismaModelName = parentType.extensions.pothosPrismaModel as string;
      const prismaRelationName = (
        connectionField?.extensions.pothosPrismaRelationField as any
      ).name as string;
      const prismaModelId = (connection as any).parent.id;
      const cache = prismaClientCache(builder as any);
      const client = cache(ctx);
      const delegate = prismaByModelName(client, prismaModelName as string);
      const result = await delegate.findFirst({
        where: { id: prismaModelId },
        select: {
          [prismaRelationName]: {
            take: 1,
            select: { id: true },
          },
        },
      });
      return result[prismaRelationName].length > 0;
    },
  })
);
```

This snippet relies on a `pothosPrismaRelationField` extension that doesn't exist yet - to obtain the original prisma relation field name. That's what's added in this PR here.

I realize that this isn't as elegant a solution as the first-class `totalCount` field, since:
- it's global for all connections regardless of whether they're prisma related connections or not
- it doesn't take advantage of the n+1 query loading aspect of the prisma plugin

If this is seen as a useful feature more broadly I'd be happy to attempt to add first-class support for an n+1 optimized existence field similar to the current `totalCount` field.

### DIFF
--- a/.changeset/hot-ligers-grin.md
+++ b/.changeset/hot-ligers-grin.md
@@ -1,0 +1,5 @@
+---
+"@pothos/plugin-prisma": patch
+---
+
+Add relationField to extensions


hayes:
This seems reasonable, we'd probably want to add the same thing for non-connection relations as well for consistency.

I am not sure about adding the existence check to the plugin directly.  If there is a good way to do this without N+1 issues, I am open to exploring it, but I don't want to introduce anything that might be a performance footgun (I don't think its possible to select the existence check and the relation in the same query)

PR #1627, opened 2026-05-14
feat(plugin-prisma-next): initial plugin for the prisma-next ORM

hayes:
## Summary

Initial version of `@pothos/plugin-prisma-next` — a Pothos plugin for the new prisma-next ORM. Stays `private: true` in the workspace until prisma-next publishes to npm.

### What's in

**Schema-builder surface**
- `builder.prismaObject` / `builder.prismaInterface` / `builder.prismaNode`
- `builder.prismaObjectField(s)` / `builder.prismaInterfaceField(s)` for cross-file extension
- `t.prismaField` / `t.prismaFieldWithInput` (top-level prisma-backed fields)
- `t.prismaConnection` (Relay top-level connection over a contract model)
- On `PrismaNextObjectFieldBuilder`:
  - `t.relation` (auto-included from the selection set)
  - `t.relatedConnection` (paginated relation under a parent type)
  - `t.variant` (re-expose the same row under a different prismaObject)
  - `t.expose*` family
  - `t.withAuth` (scope-auth interop)
- `prismaConnectionHelpers` for custom paginated fields

**Auto-include walker**
- Walks `info` and emits the orm-client chain (`.select / .include / .combine`) inline onto whatever Collection the user returns
- Honors `pothosIndirectInclude` at type and field level — interop with `plugin-errors` result-unions and `t.variant` same-row descents
- Precomputes per-type relation + column metadata in `onTypeConfig` so the per-resolve hot path never touches the contract
- Per-field overlay lifts namespaced combine slots onto flat parent properties for downstream resolvers
- Validates select shapes (typos in declarative keys throw at request time with a useful message)

**Connections**
- Cursor pagination (single or compound), forward + backward
- `totalCount` as either a synthetic count branch or a user-supplied callback, gated on whether the client actually selects it
- Edge `cursor` is a memoizing non-enumerable getter so clients selecting only `nodes` skip the per-row encode

**Resolver API**
- Resolvers return an orm-client `Collection` (or a row / row[] / null); plugin auto-detects the Collection case and applies the selection mapper + materializes via `.all()`
- Injects `.take(1)` for single-row `t.prismaField` returns to avoid full-table fetches

**Refs and brands**
- Drizzle-shape ref cache (lazy creation; variants get fresh refs)
- No auto-branding — `ref.addBrand` is exposed for the abstract-position cases that need it (matches `plugin-prisma`)
- `prismaNode` brands rows in its batch loader so the brand-check `isTypeOf` fallback works

**Cross-plugin interop**
- `plugin-relay`: `prismaNode` + `t.prismaConnection` / `t.relatedConnection`
- `plugin-errors`: result-union types descend via `pothosIndirectInclude`
- `plugin-scope-auth`: `t.withAuth` on the field-builder
- `plugin-with-input`: `t.prismaFieldWithInput`

### What's deliberately out

- M:N relations — upstream orm-client doesn't implement junction reads yet. The plugin rejects M:N at schema build with a pointer to the explicit-join-model workaround. Pinned to break the day upstream lands it (`tests/prisma-next-m-n-upstream-pin.test.ts`).
- Auto `exposeDescriptions` — the prisma-next contract doesn't carry descriptions.
- Mongo support — orm-client's Mongo Collection has a different shape; plugin is SQL-only.

### Docs

- Website docs under `website/content/docs/plugins/prisma-next/` (index, setup, objects, relations, interfaces, relay, connections, variants)
- `packages/plugin-prisma-next/ARCHITECTURE.md` — maintainer-focused walkthrough
- `packages/plugin-prisma-next/REVIEW.md` — design-review decisions
- `packages/plugin-prisma-next/feedback/prisma-team-feedback.md` — upstream-feedback summary

### Tooling

- Postinstall hook (`scripts/postinstall.mjs`) builds the sibling `prisma-next` clone if its `dist/` is missing — first `pnpm install` after a fresh clone now works without manual prisma-next build steps. Skippable via `POTHOS_SKIP_PRISMA_NEXT_BUILD=1`.
- Upstream-canary tests for the open M:N gap and Issue A (depth-2+ multi-query fallback) fail the day upstream fixes them.

## Test plan

- [x] `pnpm --filter @pothos/plugin-prisma-next test` — 201 passing, 0 type errors
- [ ] Reviewer can `git clone` Pothos + sibling-clone prisma-next, run `pnpm install` at the root, and have tests pass without manual build steps
- [ ] Docs render at `/docs/plugins/prisma-next` (verified locally)


kinglouie:
Hey @hayes, i was wondering if you could add this PR to pkg-pr-new so it is easier to test things out?

I believe M:N relations will land in the next prisma-next EA release, so hopefully that will unblock closing that feature gap :)

hayes:
@kinglouie should be published now, but the current quality bar for this is pretty low, and heavily llm generated so absolutely no guarantees about how well it works

hayes:
merged (but not published) in a separate PR

PR #1631, opened 2026-05-18
docs(enums): fix ordered list numbering and code block formatting

unrevised6419:
## Summary

- Replace repeated `1.` markers with explicit sequential numbers (1–4) for the enum definition methods list
- Dedent code blocks and explanatory prose so they render as standalone sections rather than nested list content and make syntax highlight work on the website

image

## Test plan

- [ ] Verify the enums guide page renders with correctly numbered list items and properly formatted code blocks


hayes (approved):
thanks!

PR #1634, opened 2026-05-21
fix(dataloader): exclude Error from inferred loadable Shape

### LINKED ISSUE #951: Dataloader infers wrong type from load method when returning non-base errors

ATOM00blue:
## Problem
When a dataloader `load` method resolves to `Shape | Error` (returning errors per key), the loadable refs inferred the object `Shape` without excluding `Error` from the union, so dependent field and resolver types incorrectly included `Error`.

## Fix
`loadableObject`, `loadableObjectRef`, `loadableInterfaceRef`, and `loadableNodeRef` now derive `Shape` via `ShapeFromLoadResult<LoadResult>`, which strips `Error` from the load result. `Error` values are still accepted at runtime; only the inferred object type is corrected.

## Tests
Added `packages/plugin-dataloader/tests/dataloader.test-d.ts` asserting that a `load` resolving to `{ id } | GraphQLError` infers `Shape` as `User` (Error excluded) for all four loadable ref constructors. `pnpm run type` passes.

Closes #951

### DIFF
--- a/packages/plugin-dataloader/src/global-types.ts
+++ b/packages/plugin-dataloader/src/global-types.ts
@@ -83,36 +83,77 @@ declare global {
       ) => LoadableInterfaceRef<Types, Key | Shape, Shape, Key, CacheKey>;
 
       loadableObjectRef: <
-        LoadResult,
+        Shape,
         Key extends bigint | number | string,
         CacheKey = Key,
-        Shape = ShapeFromLoadResult<LoadResult>,
+        // `LoadResult` is inferred from `options.load` and is not intended to be
+        // passed explicitly. It defaults to `Shape | Error` so that explicit
+        // generics keep working exactly as before (`loadableObjectRef<User, string>`),
+        // while inference excludes `Error` from the resolved shape (#951).
+        LoadResult = Shape | Error,
       >(
         name: string,
-        options: DataLoaderOptions<Types, LoadResult, Key, CacheKey, Shape>,
-      ) => ImplementableLoadableObjectRef<Types, Key | Shape, Shape, Key, CacheKey>;
+        options: DataLoaderOptions<
+          Types,
+          LoadResult,
+          Key,
+          CacheKey,
+          ShapeFromLoadResult<LoadResult>
+        >,
+      ) => ImplementableLoadableObjectRef<
+        Types,
+        Key | ShapeFromLoadResult<LoadResult>,
+        ShapeFromLoadResult<LoadResult>,
+        Key,
+        CacheKey
+      >;
 
       loadableInterfaceRef: <
-        LoadResult,
+        Shape,
         Key extends bigint | number | string,
         CacheKey = Key,
-        Shape = ShapeFromLoadResult<LoadResult>,
+        LoadResult = Shape | Error,
       >(
         name: string,
-        options: DataLoaderOptions<Types, LoadResult, Key, CacheKey, Shape>,
-      ) => ImplementableLoadableInterfaceRef<Types, Key | Shape, Shape, Key, CacheKey>;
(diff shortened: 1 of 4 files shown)


packages/plugin-dataloader/src/global-types.ts
```diff
@@ -82,24 +82,35 @@ declare global {
...
+      loadableObjectRef: <
+        LoadResult,
+        Key extends bigint | number | string,
+        CacheKey = Key,
+        Shape = ShapeFromLoadResult<LoadResult>,
+      >(
         name: string,
-        options: DataLoaderOptions<Types, Shape | Error, Key, CacheKey, Shape>,
+        options: DataLoaderOptions<Types, LoadResult, Key, CacheKey, Shape>,
       ) => ImplementableLoadableObjectRef<Types, Key | Shape, Shape, Key, CacheKey>;
```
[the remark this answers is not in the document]
hayes: I think this is a legit issue.

generally the builder.thingRef methods are intended to work well with explicit params,  we need to make sure that `builder.loadableObjectRef<User, string>` still allows you to return errors or this will be a breaking change that will probably break existing apps

packages/plugin-dataloader/src/global-types.ts
```diff
@@ -82,24 +82,35 @@ declare global {
...
-      loadableObjectRef: <Shape, Key extends bigint | number | string, CacheKey = Key>(
+      loadableObjectRef: <
+        LoadResult,
+        Key extends bigint | number | string,
+        CacheKey = Key,
+        Shape = ShapeFromLoadResult<LoadResult>,
+      >(
         name: string,
-        options: DataLoaderOptions<Types, Shape | Error, Key, CacheKey, Shape>,
+        options: DataLoaderOptions<Types, LoadResult, Key, CacheKey, Shape>,
```
hayes: it's a little weird, but I think this works:

```ts
options: DataLoaderOptions<Types, LoadResult | Error, Key, CacheKey, Shape>,
```

packages/plugin-dataloader/src/global-types.ts
```diff
@@ -82,26 +82,78 @@ declare global {
         >,
       ) => LoadableInterfaceRef<Types, Key | Shape, Shape, Key, CacheKey>;
 
-      loadableObjectRef: <Shape, Key extends bigint | number | string, CacheKey = Key>(
+      loadableObjectRef: <
```
hayes, line 85: I think this can be slightly simpler as:

```ts
loadableObjectRef: <
        LoadResult,
        Key extends bigint | number | string,
        CacheKey = Key,
        Shape = ShapeFromLoadResult<LoadResult>,
      >(
        name: string,
        options: DataLoaderOptions<Types, LoadResult | Error, Key, CacheKey, Shape>,
      ) => ImplementableLoadableObjectRef<Types, Key | Shape, Shape, Key, CacheKey>;
```

This basically aligns it with `loadableObject` with the only difference being you pass `LoadResult | Error` into `DataLoaderOptions` so if `LoadResult` is explicitly provided without an Error you can still load an error

hayes (approved):
Thanks for the PR, and thanks for adding tests!

PR #1647, opened 2026-08-18
[wip] fix undefined filters in Drizzle connections

williamrobertson13:
## Summary

- omit `where` from generated Drizzle connection queries when there is no filter
- cover both root `drizzleConnection` and nested `relatedConnection` query shapes
- add a patch changeset for `@pothos/plugin-drizzle`

## Why

Drizzle `1.0.0-rc.5-169397b` rejects an explicitly undefined relational filter with:

> Unexpected 'undefined' in filter value. Use 'EmptyFilter' if you want the filter field to be skipped.

Connection queries without a user filter or cursor currently include `where: undefined`. Omitting the optional property preserves existing behavior on earlier Drizzle versions and satisfies the stricter rc.5 contract.

## Validation

- regression assertions fail on both affected query paths before the fix
- `drizzle-orm@1.0.0-rc.2`: 90/90 plugin tests pass
- `drizzle-orm@1.0.0-rc.4`: 90/90 plugin tests pass
- `drizzle-orm@1.0.0-rc.5-169397b`: 90/90 plugin tests pass
- `pnpm --filter @pothos/plugin-drizzle type`
- `pnpm --filter @pothos/plugin-drizzle build`

### DIFF
--- a/.changeset/quiet-ravens-filter.md
+++ b/.changeset/quiet-ravens-filter.md
@@ -1,0 +1,5 @@
+---
+'@pothos/plugin-drizzle': patch
+---
+
+Omit undefined `where` properties from generated connection queries for compatibility with strict Drizzle relational filters.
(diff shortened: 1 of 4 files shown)


williamrobertson13:
closing this until RC 5 gets released for real

Issue #1361, opened 2024-12-26
[Bug]: Cursor-based pagination skips elements due to incorrect offset from Drizzle plugin


baristikir:
I'm using the Drizzle plugin with Relay integration and am seeing the following issue when performing a cursor-based pagination using the GraphQL Relay nodes, elements immediately following the cursor are being skipped in the results. This looks like to be caused by the Drizzle plugin incorrectly adding an offset of 1 to the query when cursor arguments are present.

### Reproduction
1. Execute a GraphQL query on either node query interface or any query returning a connection
2. Pass a cursor ID as an argument on the query for after / before 

I'm assuming that this is caused by this line:
https://github.com/hayes/pothos/blob/336f96e3b9a88375d5d15df16c8c7bebc0089058/packages/plugin-drizzle/src/utils/cursors.ts#L421

hayes:
Thanks for reporting this!

I think your right.  This looks like a remnant from porting some of this logic from prisma where cursors are inclusive. I'll try to confirm the root cause here, and get it fixed

hayes:
Should be fixed in the latest version

PR #1650, opened 2026-08-20
fix(plugin-drizzle): exclude the cursor row from keyset pages

ProductOfAmerica:
Hit this on a `last`/`before` page ordered by `{ createdAt: 'desc', id: 'desc' }`.

Date cursors are encoded at millisecond precision (`D:<epoch-ms>`, or `JSON.stringify` on a compound cursor). A default Postgres `timestamptz` is microseconds. The keyset looks exclusive:

```sql
created_at > $1 OR (created_at = $2 AND id > $3)
```

but `$1`/`$2` are the truncated value. A row stored at `.086068` is greater than `.086`, so `last: 2, before: <that cursor>` can return the same node you just named.

The extra predicate is `id <> cursor.id`, and only when the order includes a single-column number or bigint primary key. That keeps the named row out without depending on timestamp precision.

I first applied `ne` to every numeric order column. That was wrong: `category_id <> cursor.category_id` skips the rest of the current prefix on a compound keyset. Only the PK gets it now. Composite and string PKs are left alone.

This does not fix two different rows that land in the same truncated millisecond. `id` is not a stand-in for the lost microseconds, so another `id <>` predicate cannot split those neighbors. Exclusive keyset only works when the compared value equals the stored value. A JS `Date` is milliseconds; by the time the plugin sees one, the extra digits are already gone.

Two ways to close that hole, neither of which is this PR:

1. Sort at encode precision: `timestamp(3)` on the column, or `ORDER BY date_trunc('milliseconds', created_at), id`.
2. Encode at store precision: don't put a `Date` in the cursor. Drizzle `timestamp({ mode: 'string' })` keeps the raw text (including microseconds), the plugin already encodes strings as `S:…`, and Postgres casts that back to `timestamptz`.

I couldn't find an existing issue for this. #1361 looks similar (skipped rows) but it was the old `offset: 1` path and is already closed.

Changeset is included. New unit tests around `drizzleCursorConnectionQuery`, and the connection SQL snapshots updated for the extra `id <> ?`. I ran the plugin-drizzle sqlite suite locally. I did not run full `pnpm run ci`; I'm on Windows and the package `build:esm` scripts call `cp`.


ProductOfAmerica:
Closed by https://github.com/hayes/pothos/pull/1651

PR #1664, opened 2026-09-05
fix(plugin-drizzle): plan fragments on implemented interfaces, export DrizzleRef types

hayes:
## Summary

- `addNestedSelections` skipped a fragment (spread or inline) whose type condition is an interface the Drizzle object implements, because it compared the fragment's type name with the object's name. Any `select` declared on the interface's fields never reached the planned query, and the field then resolved against a row missing the relation, falling back to a separate load. The fragment is now planned against the object, whose field map already carries the interface fields and their `pothosDrizzleSelect` extensions. Fragments on other objects, and the existing interface / `pothosDrizzleModel` branch, are unchanged; the two duplicated conditions collapse into one `typeForFragment` helper.
- Export `DrizzleInterfaceRef` (as a value, like `DrizzleObjectRef` and Prisma's `PrismaInterfaceRef`, so `instanceof` works for consumers) and the `DrizzleRef` union type from the package index, so a plugin wrapping `drizzleField` can name any Drizzle ref without reaching into `interface-ref.js`.
- Changeset for a patch release of `@pothos/plugin-drizzle`.

Found while running `@pothos/plugin-drizzle` in an app where interface fields carry `select` and clients query them through `... on IWalletAddress { ... }` fragments; the same skip applies to `drizzleInterface` fields, which is what the test exercises.

## Test plan

- [x] New `tests/interface-fragments.test.ts`: `me { ... on NormalViewer { ...ViewerRoles } }` with `fragment ViewerRoles on Viewer { id roles }` (and the inline-fragment form) plans the same single query, including the `roles` relation from the interface field's `select`, as the same fragment on `NormalViewer`. Before the fix the interface form issued a bare `select id` followed by a second loader query.
- [x] `pnpm vitest --run` in `packages/plugin-drizzle`: 20 files, 120 tests pass, vitest typecheck clean
- [x] `pnpm type` in `packages/plugin-drizzle`
- [x] `pnpm build` in `packages/plugin-drizzle`; `esm/types.js` and `lib/types.js` export `DrizzleInterfaceRef`, `dts/types.d.ts` re-exports both
- [x] `biome check` on the changed files

### DIFF
--- a/.changeset/drizzle-interface-fragment-select.md
+++ b/.changeset/drizzle-interface-fragment-select.md
@@ -1,0 +1,5 @@
+---
+'@pothos/plugin-drizzle': patch
+---
+
+Plan fragments on interfaces implemented by a Drizzle object against the object, so `select` declared on the interface's fields is merged into the query instead of being skipped and loaded separately. Export the `DrizzleInterfaceRef` and `DrizzleRef` types from the package index.
--- a/.changeset/drizzle-interface-fragment-select.md
+++ b/.changeset/drizzle-interface-fragment-select.md
@@ -2,4 +2,4 @@
 '@pothos/plugin-drizzle': patch
 ---
 
-Plan fragments on interfaces implemented by a Drizzle object against the object, so `select` declared on the interface's fields is merged into the query instead of being skipped and loaded separately. Export the `DrizzleInterfaceRef` and `DrizzleRef` types from the package index.
+Plan fragments on interfaces implemented by a Drizzle object against the object, so `select` declared on the interface's fields is merged into the query instead of being skipped and loaded separately. Export `DrizzleInterfaceRef` and the `DrizzleRef` type from the package index.
--- a/packages/plugin-drizzle/src/types.ts
+++ b/packages/plugin-drizzle/src/types.ts
@@ -740,5 +740,5 @@ export interface DrizzleGraphQLInputExtensions {
   inputType: 'insert' | 'filters' | 'orderBy' | 'update';
 }
 
-export type { DrizzleInterfaceRef, DrizzleRef } from './interface-ref.js';
+export { DrizzleInterfaceRef, type DrizzleRef } from './interface-ref.js';
 export { DrizzleObjectRef } from './object-ref.js';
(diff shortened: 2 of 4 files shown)


hayes:
### Code review

Found 1 issue:

1. `DrizzleInterfaceRef` is re-exported type-only, but it is a class. The `export type` form is erased from the emitted JS, so `import { DrizzleInterfaceRef } from '@pothos/plugin-drizzle'` cannot be used with `instanceof` or `new`, which is what a plugin wrapping `drizzleField` (the stated motivation) needs to tell interface refs from object refs; the package itself does exactly that in [`utils/refs.ts#L27-L28`](https://github.com/hayes/pothos/blob/6424480487b96ee58c66934392d906b9f28bfb2b/packages/plugin-drizzle/src/utils/refs.ts#L26-L29). The precedent one line below is a value export (`export { DrizzleObjectRef }`, from #1371 resolving #1370), and the Prisma plugin value-exports the analogous class in [`plugin-prisma/src/index.ts#L22`](https://github.com/hayes/pothos/blob/6424480487b96ee58c66934392d906b9f28bfb2b/packages/plugin-prisma/src/index.ts#L21-L23). `DrizzleRef` is a plain type alias, so `export type` is right for it; only `DrizzleInterfaceRef` should move to a value export.

https://github.com/hayes/pothos/blob/6424480487b96ee58c66934392d906b9f28bfb2b/packages/plugin-drizzle/src/types.ts#L742-L744

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

hayes:
Addressed in 7e31b145ea26d02267ff9e69d1d672e357fad253: `DrizzleInterfaceRef` is now a value export (`export { DrizzleInterfaceRef, type DrizzleRef } from './interface-ref.js'`), matching `DrizzleObjectRef` and `PrismaInterfaceRef`. Confirmed in the built output: https://github.com/hayes/pothos/blob/7e31b145ea26d02267ff9e69d1d672e357fad253/packages/plugin-drizzle/src/types.ts#L742-L744. Tests, `pnpm type`, `pnpm build`, and biome all pass.

PR #1666, opened 2026-09-06
feat(plugin-prisma): support async query callbacks in relation fields

### LINKED ISSUE #1397: [Feature request]: Allow async query fields in database plugins
Following up on this discussion: https://github.com/hayes/pothos/discussions/1395#discussioncomment-12569508

it would be helpful to allow async queries in fields. In the below example a hypothetical resource server can be queried to return a database filter based on what the user is allowed to do. That way we could inject permission restrictions at query level.

```ts
const PostRef = schemaBuilder.drizzleObject("posts", {
  name: "Post",
  fields: (t) => ({
    id: t.exposeInt("id"),
    content: t.exposeString("content"),
    author: t.relation("author", {
      // filter returns a promise which resolves into a query filter depending on the context
      query: async (_args, ctx) => await ctx.abilities.users.filter("read"),
    }),
  }),
});
```

babbarankit:
## Summary

Closes #1397

- `QueryForField` return type now accepts `MaybePromise`, enabling async permission-based filtering:
  ```ts
  query: async (args, ctx) => await ctx.abilities.users.filter("read")
  ```
- `relation()` fallback path handles async query results via `isThenable` check
- `relatedConnection()` splits sync (selection tree) and async (fallback/resolve) paths — selection tree uses base pagination when query is async

## Trade-off

When `query` is async, the **batched selection tree does NOT include the filter** (selection building is synchronous). The relation IS included and data IS fetched, but unfiltered from the batch. The `resolve` fallback path applies the filter correctly by awaiting the async result.

## Test plan

- [x] All 89 existing plugin-prisma tests pass (sync behavior unchanged)
- [x] Type change is additive — `MaybePromise<T>` accepts both `T` and `Promise<T>`
- [ ] Manual: async `query` with `resolve` correctly filters results
- [ ] Manual: async `query` without `resolve` — fallback behavior

### DIFF
--- a/packages/plugin-prisma/src/prisma-field-builder.ts
+++ b/packages/plugin-prisma/src/prisma-field-builder.ts
@@ -214,8 +214,8 @@ export class PrismaObjectFieldBuilder<
     const formatCursor = getCursorFormatter(relationField.type, this.builder, cursorValue);
     const parseCursor = getCursorParser(relationField.type, this.builder, cursorValue);
 
-    const getQuery = (args: PothosSchemaTypes.DefaultConnectionArguments, ctx: {}) => {
-      const connectionQuery = prismaCursorConnectionQuery({
+    const getConnectionQuery = (args: PothosSchemaTypes.DefaultConnectionArguments, ctx: {}) =>
+      prismaCursorConnectionQuery({
         parseCursor,
         ctx,
         maxSize,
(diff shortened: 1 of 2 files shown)


hayes:
thanks for the PR!  Unfortunately I don't think this is a viable solution to the problem.  Relying on the fallback pattern is a n easy footgun that leads to very inefficient data loading in a way that is unintuitive.

I think anyone who wants something like this would be much better off just using t.prismaField/t.prismaConnection instead, which more accurately conveys that this is always going to result in a separate query.

I don't think there is a good way to support this without actually rewriting the whole planner to be async compatible which comes with a lot of its own complexity.  I am definitely open to exploring that as an option, but it would require a lot of careful design, and would probably be a good opportunity to fully rewrite/extract the "query mapping" logic so it can be shared across implementations, there are 2 now (drizzle and prisma) and a 3rd prisma next implementation in a PR which I'd probably want to move onto shared infra if I was doing a full rewrite any ways.

If you are interested in pursuing this, let me know and we can try to figure out what actually makes sense, but I'd recommend trying to get the high level design figured out and reviewed before spending a lot of time on the actual implementation

hayes:
I ended up rebuilding the query-mapping logic and adding async-support across both prisma and drizzle, which should be out in the next release

PR #1667, opened 2026-09-06
fix(core): migrate deprecated graphql-js APIs for v17/v18 compat

### LINKED ISSUE #1637: Migrate off graphql APIs deprecated in 17 / removed in 18 (defaultValue, astFromValue, scalar coercion)
## Background

graphql 17 support (#1636) deliberately keeps using a set of graphql APIs that are **deprecated in 17 and scheduled for removal in graphql 18**. Everything works correctly on 17 today — this issue tracks the migration we'll need before graphql 18, and documents one non-obvious trap so it isn't re-discovered the hard way.

All line numbers are against the `graphql-v17-support` branch.

## Deprecated APIs Pothos relies on (removed in graphql 18)

graphql 17 marks each of these `@deprecated ... will be removed in v18`:

| Deprecated | Replacement |
|---|---|
| `GraphQLScalarTypeConfig.serialize` | `coerceOutputValue` |
| `GraphQLScalarTypeConfig.parseValue` | `coerceInputValue` |
| `GraphQLScalarTypeConfig.parseLiteral` | `coerceInputLiteral` (+ `replaceVariables`) |
| `astFromValue()` | `valueToLiteral()` |
| `GraphQLInputField.defaultValue` / `GraphQLArgument.defaultValue` (JS value) | `default: { literal }` / `defaultValueLiteral` (AST) |

Note: scalar input/output coercion hooks (`coerceOutputValue`/`coerceInputValue`/`coerceInputLiteral`/`valueToLiteral`) are now accepted by `scalarType` (added alongside #1636), so user-defined scalars can already move off `serialize`/`parseValue`/`parseLiteral`. The internal usages below still need migration.

## The trap: `astFromValue` → `valueToLiteral` is NOT a drop-in swap

They consume **opposite value shapes**, verified against graphql 17.0.1:

- `astFromValue(value, type)` takes the **internal/coerced** value and serializes it.
  - `astFromValue(new Date(...), DateScalar)` → `StringValue "2020-..."`
  - `astFromValue("red", EnumColor)` → `EnumValue RED` (internal value lookup)
- `valueToLiteral(value, type)` takes the **external** value.
  - `valueToLiteral(new Date(...), DateScalar)` → garbage empty `ObjectValue {}`
  - `valueToLiteral("red", EnumColor)` → `undefined` (only the external name `"RED"` works)

Pothos stores `defaultValue` as the **internal/input** shape (`InputShapeFromTypeParam`), which is what `astFromValue` consumes. A naive `astFromValue → valueToLiteral` swap would silently corrupt custom-scalar and enum default literals.

Also note: when Pothos sets the deprecated `defaultValue`, graphql 17's **new `field.default` accessor returns `undefined`** (the two are not cross-populated). So the correct v18 path is to thread an AST literal through Pothos's config (populate `default`/`defaultValueLiteral`) rather than reconstruct it from the coerced value.

## Migration checklist (by file)

- [ ] **`packages/core`** default-value storage — `src/fieldUtils/input.ts:144,166` store `opts.defaultValue` (JS value), spread verbatim into the GraphQL config at `src/build-cache.ts:378-388` (the deprecated `defaultValue` field). Decide whether to also emit `default`/`defaultValueLiteral`.
- [ ] **`packages/plugin-directives/src/mock-ast.ts`** — highest exposure. `astFromValue(field.defaultValue, field.type)` at `:264` and `astFromValue(arg.defaultValue, arg.type)` at `:288` (plus general arg use at `:149`). Migrate using the stored AST literal, NOT `valueToLiteral(field.defaultValue)`.
- [ ] **`packages/plugin-add-graphql/src/schema-builder.ts:144,217,335`** — reads deprecated `.defaultValue` off imported graphql types; may also need to read `.default`/literal from 17-native source schemas.
- [ ] **`packages/plugin-sub-graph/src/index.ts:232,294`** — copies `argConfig.defaultValue` / `fieldConfig.defaultValue` into rebuilt configs.
- [ ] **`packages/converter/src/index.ts:663`** — reads `arg.defaultValue` for codegen (converter is graphql-16-only today; lower priority). Pre-existing minor gap: `writeInputShape()` (~473-496) does not emit `defaultValue` for input-object fields, only for args.
- [ ] **`packages/core` scalar internals** — migrate built-in/internal scalar definitions off `serialize`/`parseValue`/`parseLiteral` if/when graphql 16 support is dropped.

## Test/snapshot surface guarding default-value printing

These break first and are the regression net for the migration:

- `packages/plugin-directives/tests/index.test.ts:70-112` — asserts the `astFromValue`-produced AST nodes directly (Boolean/Enum/Int/String/List/Object).
- `= default` SDL snapshots: `plugin-directives/tests/__snapshots__/index.test.ts.snap`, `core/tests/__snapshots__/index.test.ts.snap`, `plugin-relay`, `plugin-federation`, `plugin-dataloader`, `plugin-prisma`/`plugin-drizzle`, `converter/tests/__snapshots__/index.test.ts.snap`.

## Acceptance

- No use of `serialize`/`parseValue`/`parseLiteral`/`astFromValue`/`.defaultValue` in library source.
- Default values for custom scalars and enums round-trip correctly in SDL and execution on graphql 18.
- Tests pass on graphql 16, 17, and 18.

babbarankit:
## Summary

Closes #1637

Fixes enum and custom scalar default values breaking in graphql v17+. The root cause: graphql v17 validates default values using `getDefaultValueAST()` which requires the `default: { value }` API (external values), not just the deprecated `defaultValue` (internal values).

### Key insight — the value-shape trap

`astFromValue(value, type)` takes **internal/coerced** values (e.g., `2` for enum `TWO`).  
`valueToLiteral(value, type)` takes **external** values (e.g., `"TWO"`).  
Naive swap silently corrupts enum and custom scalar defaults.

### Approach: serialize internal→external at build time

Rather than swapping `astFromValue` → `valueToLiteral`, we:
1. **Serialize** internal values to external form via `type.serialize()` at build time
2. **Emit** `default: { value: serialized }` alongside deprecated `defaultValue`
3. **Prefer** stored AST literals from v17+ `default` API in mock-ast, falling back to `astFromValue` (v16/v17) then `valueToLiteral` + serialize (v18+)

### Files changed

| File | Change |
|------|--------|
| `core/src/build-cache.ts` | Emit `default: { value: serializeDefaultValue(...) }` in `buildInputFields()` |
| `plugin-directives/src/mock-ast.ts` | Replace 3x `astFromValue()` with `internalValueToLiteral()` helper |
| `plugin-add-graphql/src/schema-builder.ts` | Read from `.default?.value ?? .defaultValue` |
| `plugin-sub-graph/src/index.ts` | Copy `default` property in config reconstruction |
| `converter/src/index.ts` | Read from `.default?.value ?? .defaultValue` |

### Backward compat

All changes are additive — graphql v16 ignores unknown config keys, v17+ uses them. No breaking changes.

## Test plan

- [x] 155 tests pass across 8 packages (core, directives, prisma, sub-graph, add-graphql, relay, federation, converter)
- [x] **Fixes 2 pre-existing test failures** in plugin-directives (enum default values in `printSchema` and `validateSchema`)
- [x] New test: validates `printSchema` produces `enumWithDefault: EN = TWO` for enum with internal value `2`
- [x] Existing test validates AST structure for enum/boolean/string/ID/list defaults

### DIFF
--- a/packages/core/src/build-cache.ts
+++ b/packages/core/src/build-cache.ts
@@ -21,6 +21,7 @@ import {
   GraphQLString,
   type GraphQLTypeResolver,
   GraphQLUnionType,
+  getNamedType,
 } from 'graphql';
 import type { SchemaBuilder } from './builder.js';
 import type { ConfigStore } from './config-store.js';
@@ -64,6 +65,48 @@ type NullableInputType =
   | GraphQLEnumType
   | GraphQLInputObjectType
   | GraphQLList<GraphQLInputType>;
+
+/**
+ * Serialize a default value from internal/coerced form to external form.
+ * Required for graphql 17+ where `default: { value }` expects external values.
+ * For enums, converts internal value (e.g., 2) to external name (e.g., "TWO").
+ * For scalars, applies the serialize function.
+ * For input objects and lists, recursively serializes nested values.
+ */
+function unwrapType(type: GraphQLInputType): GraphQLInputType {
+  return type instanceof GraphQLNonNull ? type.ofType : type;
+}
+
+function serializeDefaultValue(value: unknown, type: GraphQLInputType): unknown {
+  if (value == null) {
+    return value;
+  }
+
+  const unwrapped = unwrapType(type);
+
+  if (Array.isArray(value) && unwrapped instanceof GraphQLList) {
+    const itemType = unwrapType(unwrapped.ofType);
+    return value.map((item) => serializeDefaultValue(item, itemType));
+  }
+
+  const namedType = getNamedType(type);
+
+  if (namedType instanceof GraphQLEnumType || namedType instanceof GraphQLScalarType) {
+    return namedType.serialize(value);
+  }
+
+  if (namedType instanceof GraphQLInputObjectType && typeof value === 'object') {
+    const fields = namedType.getFields();
+    const result: Record<string, unknown> = {};
+    for (const [key, val] of Object.entries(value as Record<string, unknown>)) {
+      const field = fields[key];
+      result[key] = field ? serializeDefaultValue(val, field.type) : val;
+    }
+    return result;
+  }
+
+  return value;
+}
 
 export class BuildCache<Types extends SchemaTypes> {
   types = new Map<string, GraphQLNamedType>();
@@ -376,9 +419,18 @@ export class BuildCache<Types extends SchemaTypes> {
       const config = this.inputFieldConfigs.get(originalConfig)!;
 
       if (config) {
+        const builtType = this.buildInputTypeParam(config.type);
+
         built[fieldName] = {
           ...config,
-          type: this.buildInputTypeParam(config.type),
+          type: builtType,
+          // graphql 17+ default value API; uses 'default: { value }' with the external
+          // (serialized) value. Enum internal values (e.g., 2) must be serialized to
+          // external names (e.g., "TWO") to avoid v17 validation errors.
+          // Spread so keys don't trip graphql 16's narrower config types. Ignored on v16.
+          ...(config.defaultValue !== undefined
+            ? { default: { value: serializeDefaultValue(config.defaultValue, builtType) } }
+            : {}),
           extensions: {
             ...config.extensions,
             pothosOptions: config.pothosOptions,
--- a/packages/plugin-add-graphql/src/schema-builder.ts
+++ b/packages/plugin-add-graphql/src/schema-builder.ts
@@ -141,7 +141,8 @@ proto.addGraphQLObject = function addGraphQLObject<Shape>(
             ...input,
             description: arg.description ?? undefined,
             deprecationReason: arg.deprecationReason ?? undefined,
-            defaultValue: arg.defaultValue,
+            // biome-ignore lint/suspicious/noExplicitAny: v17+ 'default' API
+            defaultValue: (arg as any).default?.value ?? arg.defaultValue,
             extensions: arg.extensions,
             astNode: arg.astNode ?? undefined,
           });
@@ -214,7 +215,8 @@ proto.addGraphQLInterface = function addGraphQLInterface<Shape = unknown>(
             ...resolveInputType(this, arg.type),
             description: arg.description ?? undefined,
             deprecationReason: arg.deprecationReason ?? undefined,
-            defaultValue: arg.defaultValue,
+            // biome-ignore lint/suspicious/noExplicitAny: v17+ 'default' API
+            defaultValue: (arg as any).default?.value ?? arg.defaultValue,
             extensions: arg.extensions,
             astNode: arg.astNode ?? undefined,
           });
@@ -332,7 +334,8 @@ proto.addGraphQLInput = function addGraphQLInput<Shape extends {}>(
         combinedFields[fieldName] = t.field({
           ...resolveInputType(this, field.type),
           description: field.description ?? undefined,
-          defaultValue: field.defaultValue,
+          // biome-ignore lint/suspicious/noExplicitAny: v17+ 'default' API
+          defaultValue: (field as any).default?.value ?? field.defaultValue,
           extensions: field.extensions,
           astNode: field.astNode ?? undefined,
         });
--- a/packages/converter/src/index.ts
+++ b/packages/converter/src/index.ts
@@ -660,8 +660,10 @@ export default class PothosConverter {
             this.writeRequiredness(writer, arg.type);
             writer.write(',');
             writer.newLine();
-            if (arg.defaultValue != null) {
-              writer.write(`defaultValue: ${JSON.stringify(arg.defaultValue)}`);
+            // biome-ignore lint/suspicious/noExplicitAny: v17+ 'default' API
+            const defaultVal = (arg as any).default?.value ?? arg.defaultValue;
+            if (defaultVal != null) {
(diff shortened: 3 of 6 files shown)


packages/core/src/build-cache.ts
```diff
@@ -65,6 +66,48 @@ type NullableInputType =
...
 
+/**
+ * Serialize a default value from internal/coerced form to external form.
+ * Required for graphql 17+ where `default: { value }` expects external values.
+ * For enums, converts internal value (e.g., 2) to external name (e.g., "TWO").
+ * For scalars, applies the serialize function.
+ * For input objects and lists, recursively serializes nested values.
+ */
+function unwrapType(type: GraphQLInputType): GraphQLInputType {
+  return type instanceof GraphQLNonNull ? type.ofType : type;
```
hayes, line 77: Can we remove the instanceof checks throughout the pr, graphql exports safer helpers for this

packages/core/src/build-cache.ts
```diff
@@ -65,6 +66,48 @@ type NullableInputType =
...
   | GraphQLList<GraphQLInputType>;
 
+/**
+ * Serialize a default value from internal/coerced form to external form.
+ * Required for graphql 17+ where `default: { value }` expects external values.
+ * For enums, converts internal value (e.g., 2) to external name (e.g., "TWO").
+ * For scalars, applies the serialize function.
+ * For input objects and lists, recursively serializes nested values.
+ */
+function unwrapType(type: GraphQLInputType): GraphQLInputType {
```
hayes, line 76: This name is not intuitive and I'd expect it to also handle list types (which is not what we want in this case)

packages/core/src/build-cache.ts
```diff
@@ -65,6 +66,48 @@ type NullableInputType =
...
+
+  if (Array.isArray(value) && unwrapped instanceof GraphQLList) {
+    const itemType = unwrapType(unwrapped.ofType);
+    return value.map((item) => serializeDefaultValue(item, itemType));
+  }
+
+  const namedType = getNamedType(type);
+
+  if (namedType instanceof GraphQLEnumType || namedType instanceof GraphQLScalarType) {
+    return namedType.serialize(value);
```
hayes, line 95: Do other types not have an equivalent method?

Does this work across graphql versions?

packages/plugin-add-graphql/src/schema-builder.ts
```diff
@@ -214,7 +215,8 @@ proto.addGraphQLInterface = function addGraphQLInterface<Shape = unknown>(
             ...resolveInputType(this, arg.type),
             description: arg.description ?? undefined,
             deprecationReason: arg.deprecationReason ?? undefined,
-            defaultValue: arg.defaultValue,
+            // biome-ignore lint/suspicious/noExplicitAny: v17+ 'default' API
+            defaultValue: (arg as any).default?.value ?? arg.defaultValue,
```
hayes, line 219: Can we do this with a non any type cast

packages/core/src/build-cache.ts
```diff
@@ -376,9 +419,18 @@ export class BuildCache<Types extends SchemaTypes> {
...
+        const builtType = this.buildInputTypeParam(config.type);
+
         built[fieldName] = {
           ...config,
-          type: this.buildInputTypeParam(config.type),
+          type: builtType,
+          // graphql 17+ default value API; uses 'default: { value }' with the external
+          // (serialized) value. Enum internal values (e.g., 2) must be serialized to
+          // external names (e.g., "TWO") to avoid v17 validation errors.
+          // Spread so keys don't trip graphql 16's narrower config types. Ignored on v16.
```
hayes, line 430: We can probably remove these comments (or at least strip it to a small single line)

hayes (changes requested):
I started reviewing this, and  I think at the surface it looks reasonable, but I think some if the underlying concepts from my original issue where llm slop and not well thought through.

Taking a step back, I am not sure trying to force a compatibility layer is really the right scope of change for pothos to make.

There is a reason that the required helpers are being removed from graphql and they are forcing everyone into an arguable harder to use set of APIs.  Fundamentally there is no reliable/safe automatic conversion from runtime "internal" input values to the "external" representations that introspection/schema sdl depend on (and what graphql 18 uses).  The mapping is explicitly one directional (external -> internal) and not intended to reversible. 

We can get a reasonable approximation (this is what graphql 17 does today), but its fundamentally unsound.

I think it might be better to just align Pothos with graphql-js instead and infer the available options based on the installed graphql-js version and proxy the values through without any internal mappings:

* defaultValue: available in 16/17 not in 18
* default: { value | literal } in 17/18

Making this work in the type system is pretty easy without a breaking change, and the breaking change in 18 remains the domain of graphql-js not pothos.  We probably don't need runtime enforcement, and would just pass the values through unchanged to the input field/argument options.

The hard piece here would be extending the `default` option with proper typing for values that are type-safe.  We don't have a representation of types "external" values which we'd need to thread through scalars, enums, and input types.  I think this is doable by mirroring some of the existing patterns. I'm not sure how much of this can be supported cleanly without breaking changes though.

Let me know if you are interested in continuing to work on this.  I think this is probably a significantly different scope of work than whats in this PR, but I think it would put pothos on much more stable footing in the future by aligning more closely with how things are intended to work in graphql js

PR #1669, opened 2026-09-07
fix(prisma, drizzle): resolve indirect include paths through fragments

### LINKED ISSUE #1668: `plugin-prisma`: are `{ name, type }` path segments supported by `queryFromInfo`? Types say `string[]`, runtime normalises objects
### Package

`@pothos/plugin-prisma` 4.15.0

### Question

`queryFromInfo` declares its `path` / `paths` parameters as `string[]` / `string[][]`, but the implementation normalises each segment as though the richer `{ name, type }` form were also valid input:

https://github.com/hayes/pothos/blob/main/packages/plugin-prisma/src/util/map-query.ts

```ts
path?: string[];
paths?: string[][];
```

```ts
paths.length > 0
  ? paths.map((p) => p.map((n) => (typeof n === 'string' ? { name: n } : n)))
  : [path.map((n) => (typeof n === 'string' ? { name: n } : n))]
```

Under the declared `string[]`, the `: n` branch is unreachable — `n` narrows to `never` — so I wanted to check whether accepting objects here is intended and the declaration is just narrower than it should be, or whether that normalisation is leftover from an earlier signature and shouldn't be relied on.

A couple of things suggest the object form is the intended currency rather than an accident:

- `resolveIndirectInclude` and `resolveIndirectIncludePaths` are both declared as `{ type?: string; name: string }[]`.
- The sibling API is typed for both forms — `FieldSelection`'s `mergeNestedSelection` takes `path?: IndirectInclude | string[]` — and the plugin's own connection code passes the rich form through it (`prisma-field-builder.ts`, `paths: [[{ name: 'nodes' }], [{ name: 'edges' }, { name: 'node' }]]`).

### Why it matters

The `type` on a segment isn't cosmetic when the target field sits behind an inline fragment. In `resolveIndirectInclude`, the `INLINE_FRAGMENT` branch only carries `expectedType` forward when `include.type` is set:

```ts
include.type ? info.schema.getType(include.type) : expectedType
```

and the `FIELD` branch then gates on `expectedType.name === type.name`. Without the `type`, `expectedType` stays as the outer type while `type` becomes the fragment's type condition, the names don't match, and the field is skipped — so the selection isn't merged and the plugin falls back to a `findUnique` per row.

Concretely, I have a calendar query returning an interface (`SpaCalendarEntry`, with a `resolveType`) implemented by two object types. Only one of them carries the Prisma-backed field, so it is selected behind an inline fragment:

```graphql
lanes { entries { ... on SpaCalendarAppointmentEntry { appointment { ... } } } }
```

The path that works at runtime is:

```ts
queryFromInfo({
  context,
  info,
  typeName: 'SpaBookingGuestTreatment',
  path: [
    { name: 'lanes' },
    { name: 'entries' },
    { name: 'appointment', type: 'SpaCalendarAppointmentEntry' },
  ],
  include: { appointment: { select: { roomID: true, staffID: true } } },
})
```

It produces the correct single query, and I have a test asserting the per-row `findUnique` fallback never fires. But it only compiles behind `as unknown as string[]`, since `{ name, type }[]` and `string[]` aren't comparable.

### Ask

This may just need clarification rather than a code change. Is it something that can be fixed — widening `path` / `paths` so the object form type-checks — or have I misunderstood how this is meant to be used?

If I've misunderstood, a pointer to the intended way to target a field behind an inline fragment would be welcome; I couldn't find one in the docs, and dropping the `type` loses the selection. Otherwise, please just comment and close.

hayes:
## Summary

Closes #1668.

`queryFromInfo` and `nestedSelection` follow a `path` through the query's selection set to find where a type's selections live. Three things were wrong in how that walk handled fragments, in both `@pothos/plugin-prisma` and `@pothos/plugin-drizzle`:

- **Narrowing fragments were skipped.** Since the `totalCount` fix in 0be6da5c, prisma's `resolveIndirectInclude` only advanced the type it expected a segment on when the segment carried an explicit `type`. A field selected behind `... on Impl { field }` under an interface or union list never matched, so the selection was not merged and the plugin fell back to a `findUnique` per row. The reporter worked around it with `{ name, type }` segments behind an `as unknown as string[]` cast. Drizzle's copy of the walker predates that change and handled inline fragments, but crashed on a fragment *spread* that narrows an interface.
- **`nestedSelection` broke through any fragment.** `normalizeInclude` recorded each segment's *return* type as its `type`, while the walker compares a segment `type` against the enclosing fragment's *type condition*. Those are different types, so `nestedSelection(q, ['post'])` produced an empty selection (and a Prisma "select must not be empty" error) whenever the field was reached through `... on PostPreview { post }` or `...Fragment`, even on the identical type. This is independent of the issue and has been latent since the helper was added in 2022; it only stayed hidden because `nestedSelection` with a non-empty path is rare.
- **The public type was narrower than the runtime.** `path`/`paths` were declared `string[]` while the implementation normalised `{ name, type }` objects since the option was introduced.

## Changes

- The walker is now `findIndirectSelections` (both plugins): it returns every field matched by a set of paths instead of invoking a callback per match, and replaces the old `resolveIndirectInclude` / `resolveIndirectIncludePaths` pair. Fragments are classified against the expected type by `resolveFragmentTypes`: an explicit segment `type` pins it, a fragment on the expected type or on an abstract type it belongs to walks as the expected type, a fragment that narrows an abstract expected type (a declared implementation or member, or an abstract type that overlaps it per `doTypesOverlap`) advances it, and an unrelated fragment is skipped until a nested fragment narrows back. Both fragment kinds behave the same. An unknown segment `type` throws a `PothosValidationError`. The `totalCount` split-fragment case from 0be6da5c still passes.
- Every match is merged into one selection state. Previously `queryFromInfo` created a fresh state per match so only the last one survived, which also dropped selections when both alternatives of `paths` (e.g. `edges.node` and `nodes`) were selected.
- A matched field whose type has its own model, including a variant of the target model, is walked with its own field map. Walking a variant as the requested type threw on variant-only fields.
- `normalizeInclude` (both plugins) emits `{ name }` segments only, so string paths are matched by name and narrowed by the walker rather than pinned to a type the walker never meant.
- Because a string path through an interface list now matches every implementation with that field name, `findIndirectSelections` takes a `targetType` and drops matches whose field returns a different prisma model / drizzle table. Types without a model (interfaces, wrappers) always pass. Without this, two implementations with a same-named field of different types went from silently falling back on `main` to throwing `Unknown field ... on User` (prisma) or a missing table config error (drizzle). `{ name, type }` segments remain the explicit way to restrict a match to one implementation.
- The type-level indirect include check in `addTypeSelectionsForField` used `paths.length === 0` where it meant `> 0` (unreachable today, only `path` is ever produced).
- `queryFromInfo` `path`/`paths` accept `(string | { name: string; type?: string })[]`. Drizzle's internal `QueryFromInfoOptions` gets the same shape for consistency.
- Docs: a short note in the prisma selections page on how `path` follows fragments and when to pin a segment `type`.
- Changeset: patch for both plugins.

## Test plan

- [x] New `packages/plugin-prisma/tests/indirect-include-fragments.test.ts` (14 tests): `queryFromInfo` path through an inline fragment, a fragment spread, a narrowing-then-widening fragment pair, a fragment on an overlapping interface, a union member, and a `{ name, type }` segment, each asserting a single query with the include and the expected data; a same-named `appointment` field on two implementations returning `User` and `Profile`, asserting no error and only the `User` query; a `User` variant on a third implementation walked with its own fields and merged; a pinned segment skipping the variant's selection; two fragments selecting the same field merged into one include; an unknown segment `type` reported as a validation error; `nestedSelection` path selected directly, through a same-type inline fragment, and through a same-type fragment spread, each asserting the identical Prisma query. Five of the original seven fail on `main`; the different-model case throws without the `matchesTargetModel` guard.
- [x] New `packages/plugin-drizzle/tests/indirect-include-fragments.test.ts` (10 tests) mirroring the above, including the different-table, variant, overlap, union, merge and validation cases, with a drizzle schema and comparing planned SQL. The fragment-spread and `nestedSelection` cases fail on `main`.
- [x] `pnpm vitest run` in `packages/plugin-prisma`: 15 files, 103 tests, typecheck clean
- [x] `pnpm vitest run` in `packages/plugin-drizzle`: 21 files, 130 tests, typecheck clean
- [x] `pnpm type` in both packages (the first CI run failed here on type errors in the new tests: string type names in the prisma test and a required `scopeAuth` option in the drizzle test; both fixed in the second commit)
- [x] `biome check` on changed files


hayes:

PR #1670, opened 2026-09-08
fix(prisma, drizzle): relation selection and loading bugs

hayes:
Stacked on #1669. Second PR in the selection-mapper stack; each commit is one fix with its own test.

## Prisma

- **`where: undefined` no longer splits a relation selection.** A relation `query` returning `{ where: cond ? filter : undefined }` was incompatible with a sibling selection of the same relation whenever the condition was false, so the sibling fell back to a per-row query. `deepEqual` now treats keys holding `undefined` as absent on both sides. (`tests/relation-options.test.ts`)
- **`onNull` receives `args`, `context`, `info`.** It was called with empty objects, so any callback reading them threw. (`tests/relation-options.test.ts`)

## Drizzle

- **`relatedConnection` `totalCount` applies the field's `where`.** The count was built from the relation columns only, so a filtered connection reported the unfiltered count; the package's own snapshot asserted the wrong number. New builder option `drizzle.filterConnectionTotalCount` (default `true`) restores the old count when set to `false`, matching prisma's option. (`tests/related-connection.test.ts` snapshots: count 15 → 9 published posts)
- **`relatedConnection` rejects `resolve` in its types.** It was accepted and silently discarded. (`tests/relation-loading.test.ts`)
- **`drizzleConnectionHelpers(builder, table)` works without an options argument**, as the docs show. It threw on destructuring. (`tests/relation-loading.test.ts`)
- **Relations missing from a parent row are loaded instead of answered from missing data.** When a resolver returned a row it fetched itself, `t.relation` returned `undefined`, `t.relatedCount` returned `undefined`, and `t.relatedConnection` returned an empty page, because the loader mapping alone was trusted. Fields now declare a `pothosDrizzleLoaded` check mirroring prisma's `pothosPrismaLoaded`, and `wrapResolve` falls back to the model loader when the data is not on the row. `t.relatedField` (and so `t.relatedCount`) now forwards every option and merges `extensions`. (`tests/relation-loading.test.ts`)

## Docs

Corrections from the API audit: the prisma pagination-combination list was inverted; the "Extending connection edges" example returned a `select` key Prisma rejects; strict-null examples; drizzle `orderBy` prose contradicted its own examples; `drizzleConnectionHelpers` and `drizzleObjectField(s)` examples passed GraphQL type names where table names are required; `totalCount` filtering and the new option; the shape of `pathInfo`.

## Not in this PR

The drizzle `relatedConnection` `pathInfo` mismatch between the select path and the resolve path is deferred to the shared-walker PR, where the mapping record that carries it is redesigned.

### DIFF
--- a/.changeset/selection-mapper-bugs.md
+++ b/.changeset/selection-mapper-bugs.md
@@ -1,0 +1,20 @@
+---
+'@pothos/plugin-prisma': patch
+'@pothos/plugin-drizzle': patch
+---
+
+Fix several bugs in how relation selections are mapped and loaded.
+
+Prisma:
+
+- A relation `query` returning keys set to `undefined` (`{ where: cond ? filter : undefined }`) is now treated as compatible with a sibling selection of the same relation. Previously the sibling fell back to a separate query per row whenever the condition was false.
+- `onNull` now receives `args`, `context`, and `info`. Previously it was called with empty objects, so any callback reading them threw at runtime.
+
+Drizzle:
+
+- `totalCount` on `t.relatedConnection` now applies the `where` returned by the field's `query`, so it counts the same rows the connection paginates. Previously it reported the unfiltered count.
+- New builder option `drizzle.filterConnectionTotalCount` (default `true`). Set it to `false` to keep counting every related row regardless of the filter, matching the prisma plugin's option.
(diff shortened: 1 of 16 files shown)


hayes:

PR #1671, opened 2026-09-08
fix(prisma, drizzle): apply the approved selection-walker behaviour rulings

hayes:
Stacked on #1670. Third PR in the selection-mapper stack. Each commit is one of the behaviour rulings agreed for the walker, applied to today's code with tests in both plugins, so that the later extraction into a shared package can be verified as "no snapshot changes". Where an existing snapshot changed, the rule that requires it is named. The last seven commits address the first review round.

## Rulings

1. **Loader fallback loads the parent type's selection.** The model-loader fallback substitutes the loaded row for the parent, so the row now carries the parent type's type-level `select`/`include` (prisma) or `select` (drizzle) plus the field's own selection. The field is merged first; a type-level relation or count whose arguments conflict with the field's own selection is left out (per entry), since the row exists to answer that field. Snapshots: prisma `defer`/`connections` fallback queries gain the example `User`'s type-level `include`; prisma `select-mode` gains `id`; drizzle fallbacks gain the type-level columns. Tests: `loader-fallback.test.ts` in both.
2. **Fragments on same-model variants and non-model interfaces.** Prisma no longer skips fields under a fragment on a non-model interface (drizzle already handled it), and both plugins classify nested fragments against the current type even under a non-applicable outer fragment. A fragment that enters a same-model variant merges that variant's type-level selection; an include-mode variant under a select-mode interface loads the whole row. Variants are entered in a first pass over each selection set, before any field at that node is merged, so the outcome does not depend on document order: a field-level selection that conflicts with a variant's type-level selection falls back, and two conflicting type-level selections throw a `PothosValidationError` naming both types and the relation (or, in drizzle, the extra), pointing at a field-level `select`. Fragments disabled by `@skip`/`@include` are not entered or planned. Prisma enters a same-model *object* variant only when the field's declared type is abstract, so `... on Viewer { ... on OtherViewer { … } }` under a concrete `NormalViewer` field is left alone. Snapshot: prisma `abstract-types` "select union" plans `author` in include mode because `... on User` enters the include-mode `User` variant. Tests: `variant-fragments.test.ts` in both.
3. **Path-bearing wrapper types keep their own selection only for their own model.** A type carrying `pothosIndirectInclude` with paths is still walked as itself after its matches, but only when it is backed by the same model as the node. Tests: `indirect-include-wrappers.test.ts` in both.
4. **`queryFromInfo` with no matching path returns the caller's selection**, or `{}`, never `{ select: {} }`. Snapshot: prisma `counts` connection-in-fragment tests keep the cursor `select: { id: true }` that `prismaConnection` passes as the initial selection. Tests: `query-from-info-no-match.test.ts` in both.
5. **Node getter returns the first match and resolves through the return type's wrapper.** A `relatedConnection` with `totalCount: true` wrapped by the errors plugin now plans the count. Prisma's planner-side and resolve-side "what does this connection need" checks use the same path lookups drizzle uses (`totalCount`, `edges`, `nodes`, `pageInfo`, across every fragment), so a connection selected across two success fragments plans both the count and the rows. Tests: `errors-wrapped-connection.test.ts` in both.
6. **A falsy field `select` selects nothing** instead of flipping the type to include mode (prisma) or recording a mapping for data that was never loaded (drizzle). Tests in `loader-fallback.test.ts`.
7. **drizzle: `pathInfo.path` starts with the root field for queries planned through `paths`.** Test: `indirect-query-path.test.ts`.

## Known follow-up (out of scope here)

When an object inherits a field from two model interfaces, the loader mapping is recorded under the walked interface but looked up under the interface that owns the field config, so the field falls back to the loader. Pre-existing; surfaced while writing the order-independence test.

## Verification

prisma: 129 tests, `pnpm type` clean. drizzle: 163 tests, `pnpm type` clean. Biome clean. No `.snap` file changed on top of the base; every inline snapshot change is listed above.


hayes:

PR #1672, opened 2026-09-09
refactor(prisma, drizzle): plan selections with a shared @pothos/selection-mapper package

hayes:
Stacked on #1671. Fourth PR in the selection-mapper stack.

The GraphQL-selection-to-query walker that both plugins carried as two mirrored ~800-line `map-query.ts` files (plus mirrored `selections`, `loader-map`, `deep-equal`, `usage`) now lives once in a new workspace package, `@pothos/selection-mapper`. Each plugin keeps an `Adapter` object (how to find a model, what a type and a field select, how to merge, compare and serialize its own map format), its field builders, model loader, `wrapResolve`, cursor and connection helpers. Built to the approved design document with the deltas agreed in review; no async in this PR (`finish` is synchronous; the structure leaves room for it).

## Behaviour changes in this PR (everything else is verified unchanged)

- **L-1: loader mapping keys carry the full relative path** (`Type@edges.node.author`). `nodes.author` and `edges.node.author` are both recorded, so neither falls back. One existing expectation changes: prisma `findUnique.test.ts` goes from 18 to 17 queries, because both `WithCompositeUniqueNode` and `WithCompositeUniqueNodeCustom` now get their `relations` mapping under `nodes` where the alias-keyed record kept only the first fragment's type. Test: `e38acf9c` in both plugins.
- **E-3: a connection selecting only `pageInfo`** loads just the cursor columns (prisma). Test: `0a9f97fc`.
- **drizzle `relatedConnection` resolves with the `pathInfo` it was planned with.** The select path recorded a mapping and now also the `PathInfo` computed for the field; the resolve path reads it back instead of calling `getQuery` without it. Test: `e84f9ec2`.
- **prisma `_count: true`** in a type-level or field-level selection is kept (it was dropped by `mergeSelection`, which only recorded `_count.select` entries). Test: `19720594`.

No `.snap` file changed. `packages/plugin-drizzle/tests/selections.test.ts` is rewritten against the drizzle adapter (same seven cases) because the `createState`/`mergeSelection`/`selectionToQuery` API it tested is gone.

## Points for the maintainer's eye

- **Adapter members beyond the design's eight** (plus `callbackExtra`): `typeLevelConflict` (S-7 error needs the offending relation or extra's name), `withoutConflicts` (E-2 per-entry filtering of the parent's selection), and optional `fragmentType` (the two plugins' fragment-classification rules stay per plugin, as agreed until you rule on unifying them). Plus an entry option `noMatch` so drizzle keeps returning the raw `select` when paths match nothing.
- **`Mapping = { nested, extra? }`** rather than a bare leaf, which is what carries the drizzle `PathInfo` to the resolve path.
- Three small divergences from the old code that no test pins, listed so they are deliberate: untyped inline fragments inherit suppression (the design's rule; the old prisma code re-classified them as the walked type); a resolved type with no model throws `Expected X to have a model` (old prisma tolerated it silently, old drizzle threw); prisma keeps `deferred` on wrapper matches like drizzle does.
- Prisma `onTypeConfig` now precomputes the type-level selection and propagates the field map to implementing types; drizzle propagates the table config. This is what lets `enter` allocate nothing.
- Sync-path allocations (reasoned): zero promises; per entry point one `Env`, one `Walk`, one `Node`; per walked type nothing; per function-select field one key string, one mapping record, two closures, the args object; one small array per selection set containing fragments, which a loop form could remove.

## Verification

`@pothos/selection-mapper`: 50 tests, `pnpm type` clean. prisma: 134 tests, `pnpm type` clean. drizzle: 167 tests, `pnpm type` clean. Root `pnpm run ci`: 106/108 turbo tasks; the failure is `plugin-prisma-utils#test` needing its Postgres container, unrelated. Biome clean over `packages examples website .changeset`.

## Review round 1 fixes (last seven commits)

- **Document order restored** in the field pass; only the variant entry is hoisted into the first pass. `... on User { first: posts(take: 1) }` followed by `second: posts(take: 2)` plans `first` again, as before. Tests: `document-order.test.ts` in both plugins and the package.
- **Every `info.fieldNodes` entry is planned** at the root entry (direct and `paths`), so a variant or field selected in a later occurrence of a merged response key is loaded, in either order. Variants of all nodes are entered before any field is merged. Tests: `variant-fragments` and new `merged-field-nodes.test.ts` in both plugins.
- **Visited set** for the variant pre-pass, the field pass and `findMatches`, keyed on fragment plus walk state, so a fragment DAG with repeated spreads is expanded once per state. Package test: depth-12 double-spread chain plans in well under a second.
- **One traversal per connection selection** at resolve time: `selectedFieldNames(context, info)` memoised per execution (keyed on the field nodes, the execution's variable values and the return type; graphql 17 allocates a fresh `fieldNodes` array per resolve, so the array cannot be the key). The loaded-check hooks now receive `context`. Tests: `connection-selection.test.ts` in both plugins.
- **`LoaderMappings`** kept in prisma as a deprecated alias of the shared `Mappings`.
- **Package surface trimmed** to what the plugins import plus the `Adapter` contract and entry points; README states the package is an implementation detail of the ORM plugins.
- `finish()` typed without the assertion and marked as the async seam.

selection-mapper 60 tests, prisma 142, drizzle 174; `pnpm type` clean in all three.

## Review round 2 fixes (last two commits)

- **Path matches enter their variants before any field is merged.** The `paths` entry (and the nested callback with a path or an explicit type) walked each match on its own, so a later match's variant could replace relation arguments after an earlier match's mapping was recorded. All matches now go through the same two-phase walk as multiple field nodes. Tests: package `walk.test.ts` and `indirect-include-fragments.test.ts` in both plugins, both match orders.
- `LoaderMappings` deprecation note no longer points at the internal package.

selection-mapper 61 tests, prisma 143, drizzle 175; `pnpm type` clean in all three.


hayes:

PR #1673, opened 2026-09-09
feat(prisma, drizzle): async selection callbacks with a zero-cost synchronous path

hayes:
Stacked on #1672. Fifth PR in the selection-mapper stack.

The three user callbacks the walker runs, argument mappers, a field's `select`, and a relation `query`, may now return promises. The mechanism is the one from the approved design (§5): one optional `pending` promise on a walk, only merges are chained onto it (never user code), and `finish` returns synchronously when nothing is pending. A prisma `select: async () => ...`, which compiled and was silently dropped, now works.

## What changed

- **Package.** `Walk.pending`, `chain`, `finish`; two chain sites (`applyField` for async args or an async select result, `nestedSelectionFor` for an async relation `query`). Callbacks run as soon as their own inputs resolve; async merges land after every synchronous merge, in append order. `relation()` still throws when a relation value is a thenable, telling the user to `await nestedSelection(...)`.
- **Every consumer** goes through a maybe-promise: prisma `t.relation`, `relatedConnection` (`getQuery`, select, loaded-path `resolve`, fallback), `relationCount` `where`, `prismaConnectionHelpers.getQuery`, the `prismaField`/`prismaConnection` resolvers, the `wrapResolve` fallback, node loading, and the model loader (caches the maybe-promise walk per `Type@path`); drizzle `t.relation`, `relatedConnection`, `relatedField`, `relatedCount`, `drizzleConnectionHelpers`, cursor connection resolution, and the model loader.
- **Types.** Inputs widened to `MaybePromise` (relation `query`, count `where`, connection-helper `select`/`query`, field `select`). Declared returns of `queryFromInfo`, `nestedSelection`, the drizzle resolver-side `query()` and the connection helpers' `getQuery` stay synchronous (A-7): in an async schema they hand back a promise the user awaits. Type tests pin both.
- **Docs.** One "Async selections" section per plugin page, including the one hole the design accepts: a `nestedSelection` whose promise is discarded leaves its selection unloaded and is not detected.

## Synchronous path

Zero promises and zero closures beyond the base, asserted by spies: over the package's fake adapter for every entry point, and over a real prisma and drizzle document through the built schema's resolvers (root field plus the loaded-path relation, count, related-field and connection resolvers for every row), with zero queries issued. Per-call sites use a split form (`isThenable(v) ? v.then(fn) : fn(v, ...)` with `fn` hoisted) rather than a closure-capturing `completeValue`. `finish` has a fixed arity, which also removes the rest-array allocation the base carried.

## Deviations from the design, deliberate

- Connection-helper `getQuery` keeps its synchronous declared return, like `nestedSelection`; widening it would break every synchronous `select` that embeds it.
- Prisma `relatedConnection`'s loaded-path `resolve` keeps reading the user `query` for `take`, since a user query can override it and the loaded path must use the same value the plan used.
- `chain` joins links with `Promise.all` from the second link on, so a callback rejecting before an earlier link settles is never an unhandled rejection.

## Verification

selection-mapper 75 tests, prisma 155, drizzle 188; `pnpm type` and biome clean in all three. Root turbo `build generate test type` 108/108 excluding `plugin-prisma-utils` (needs its Postgres container).

## Review round 1 fixes (last seven commits)

- **drizzle: the plan is settled before a `drizzleField`/`drizzleConnection` resolver runs**, so the `query()` builder the resolver receives never returns a promise. Before, an async selection anywhere beneath made `query({ where })` return a promise that user code handed straight to drizzle, which ran the query without the predicate. The package exports `queryFromWalk` for this; `walkFromInfo` returns `undefined` on no match. Tests assert the issued SQL keeps the predicate and equals the synchronous twin.
- **A select that discarded a pending `nestedSelection`** is refused with a validation error naming the field, and its rejection is handled, so no mapping can claim data the parent never loaded and nothing escapes as an unhandled rejection. Pending children are counted on the invocation record only while async work is outstanding.
- **A synchronous throw after an async sibling** no longer orphans that sibling's rejection; every entry point handles pending merges before rethrowing.
- **Model loaders stage a synchronous selection without awaiting it** (prisma and drizzle, plus prisma node loading), so the per-tick batch is intact and the resolver's synchronous window contains only the loader's own batch promises. Spy tests now cover the unloaded-row paths (relation, count, connection, node load) and batching of several raw parents.
- Prisma's exported `FieldSelection` type widened; drizzle `selectionOf` hoisted to module level; docs example awaits `queryFromInfo`; helper callbacks listed among the async-capable ones.

One non-snapshot expectation changed: planning now runs before a `drizzleField` resolver whether or not it calls `query()`, as `prismaField` already did, so a `relatedConnection` `query` callback observes the root path once more when the row later loads through the loader (data and query count unchanged).

selection-mapper 87 tests, prisma 159, drizzle 196; `pnpm type` and biome clean in all three.

### DIFF
--- a/.changeset/selection-mapper-async.md
+++ b/.changeset/selection-mapper-async.md
@@ -13,7 +13,8 @@ Allow the callbacks that build a selection to be async.
   `select` and silently dropped it.
 - The plugin still builds one query: callbacks start in the same tick, and what they return is
   merged after every synchronous selection, in document order. Schemas without async callbacks are
-  unaffected: no promise is created until a callback returns one.
+  unaffected: until a callback returns a promise, planning and resolving create no promise and no
+  closure they did not create before.
 - Inside an async `select`, `await` the result of `nestedSelection` (and of `getQuery` from the
   connection helpers) before adding it to the selection; a selection holding the promise itself
   throws with a message naming the relation. `queryFromInfo` (prisma) and the `query()` builder
--- a/.changeset/selection-mapper-async.md
+++ b/.changeset/selection-mapper-async.md
@@ -17,7 +17,8 @@ Allow the callbacks that build a selection to be async.
   closure they did not create before.
 - Inside an async `select`, `await` the result of `nestedSelection` (and of `getQuery` from the
   connection helpers) before adding it to the selection; a selection holding the promise itself
-  throws with a message naming the relation. `queryFromInfo` returns a promise when a callback
+  throws with a message naming the relation, and a `select` that returns while a nested selection
+  it started is still pending throws with a message naming the field. `queryFromInfo` returns a promise when a callback
   beneath the field is async, and must then be awaited; its declared type stays synchronous. The
   `query()` builder handed to a `drizzleField` or `drizzleConnection` resolver never returns a
   promise: the plan is settled before the resolver runs, and the builder merges into it.
--- a/.changeset/selection-mapper-async.md
+++ b/.changeset/selection-mapper-async.md
@@ -17,6 +17,7 @@ Allow the callbacks that build a selection to be async.
   closure they did not create before.
 - Inside an async `select`, `await` the result of `nestedSelection` (and of `getQuery` from the
   connection helpers) before adding it to the selection; a selection holding the promise itself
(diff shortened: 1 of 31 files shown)


hayes:

PR #1674, opened 2026-09-09
feat(prisma, drizzle): honest selection types, helper types, and docs

hayes:
Stacked on #1673. Sixth and last PR in the selection-mapper stack. Developer experience only: types and docs, no runtime enforcement.

## Type helpers (each pinned by a type test in `selection-types.test-d.ts` in both plugins)

1. **`queryFromInfo` is typed by what it returns.** With `select`/`include` the corresponding shape; otherwise a `{ select?; include? }` that spreads into a Prisma call. The example schema's cast around it is gone. Caveat documented: `select` on an include-mode type still yields `include` at runtime.
2. **`nestedSelection` is typed as the query it returns**, not the identity of its argument, and its argument is constrained to the field's relation query when the field's type names a model or table. Unknown keys still pass, so every existing usage compiles. The connection helpers' own `select` callback stays identity-typed: the node's model is the field's `type`, which the helper cannot know, and typing it broke the documented indirect-relations pattern.
3. **One `PathSegment` type** (`string | { name; type? }`) shared by `nestedSelection`, `queryFromInfo` and `IndirectInclude`, exported from the package and re-exported by both plugins. A runtime test shows a typed segment restricting the plan to the pinned implementation.
4. **A relation's fallback `resolve` receives a `query` typed with the relation's own keys** (`where`, `orderBy`, `take`, ...). The two casts in the example schema are gone.
5. **`relationCount` / `relatedCount` reject to-one relations** in the types.
6. **`t.relatedField` accepts the ordinary field options** (`deprecationReason`, `extensions`, `authScopes`, ...), `resolve` may return a promise, `info` is `GraphQLResolveInfo`.
7. **`PrismaQueriedShape<Types, Model, Query>` / `DrizzleQueriedShape<Types, Table, Query>`**: opt-in row-origin helper types for users who wrap rows themselves. Pure types.
8. **Drizzle's dead `withUsageCheck` plumbing is deleted.** Prisma's `onUnusedQuery` wraps a query object in getters; a drizzle resolver is handed a builder function, so that mechanism cannot see the realistic mistake (never calling `query()`), and a call-tracking equivalent is out of this PR's scope. The changeset says so.

## Docs

Each plugin's selections page gains a "how selections are planned" section: a field is either planned into an ancestor's query and mapped, or loaded by its own resolver or the loader, never both; two selections share a node only when their arguments match; `select` merges with siblings. Everything the audit found undocumented is now documented: `findUnique: null`, `onNull`, `select` on a nested `t.prismaField`, typed path segments and `nestedSelection`'s third argument, `skipDeferredFragments`, the `*FieldWithInput` methods, drizzle's connection size options, `drizzleInterface` and `drizzleInterfaceField(s)`, `t.variant(Ref, { select })`, the drizzle fallback loader, and the conflict validation error. Doc examples where the code won: `prisma.city.create` → `prisma.post.create`, a bare `nullable:` → `nullable: true`, and an unclosed code block at the end of the drizzle page.

## Verification

selection-mapper 76 tests, prisma 160, drizzle 191; `pnpm type` and biome clean (no warnings) in all three.

## Review round 1 fixes (last ten commits)

- `PrismaRelationQuery` gains `omit` and `distinct`, matching the generated relation args (the generated `Parent$relationArgs` type is not reachable from `PrismaModelTypes`, so the keys are spelled out and pinned against `Prisma.User$postsArgs`).
- String-named prisma object types (`type: 'Post'` with `Objects: { Post }`) get the same model-typed `nestedSelection` as refs.
- `PrismaQueriedShape` models `_count` the way Prisma returns it, for `include` and for `select` with `true` or a selection.
- `nestedSelection` accepts the callback and promise forms the runtime always accepted.
- `queryFromInfo({ select })` is typed as the union of the select and include shapes, since the type's mode wins at runtime; spreading into `findMany` still compiles.
- drizzle `drizzleInterfaceField(s)` now resolves a table name to the interface ref, as the docs promised (runtime fix with a test).
- Changeset corrected (`IndirectInclude` paths take object segments only) and docs qualified: a nested field with a `select` is planned into the parent and still runs its own query; type-level selections are planned before fields; the drizzle loader falls back to the first unique column for a table without a primary key.

## Open decision for the maintainer

`relationCount` / `relatedCount` now reject to-one relations at the type level. Code calling them on a to-one relation compiled before and failed at runtime with a query Prisma or drizzle rejects. The changeset says so plainly and keeps the `patch` bump; whether that should be a larger bump is your call.

selection-mapper 88 tests, prisma 164, drizzle 201; `pnpm type` and biome clean in all three.

### DIFF
--- a/.changeset/selection-mapper-types.md
+++ b/.changeset/selection-mapper-types.md
@@ -1,0 +1,40 @@
+---
+'@pothos/selection-mapper': patch
+'@pothos/plugin-prisma': patch
+'@pothos/plugin-drizzle': patch
+---
+
+Type the selection API by what it does, and document what was undocumented.
+
+- `queryFromInfo` (prisma) is typed by what it was given: `{ select }` or `{ include }` when one
+  was passed, and otherwise `{ select?, include? }`, whichever the type's mode produced. It used
+  to claim `{ select: Select }` for every call without `include`.
+- `nestedSelection` returns the relation query for the field's model or table (prisma:
+  `PrismaRelationQuery<Model>`; drizzle: the table's `DBQueryConfig`, a `many` config for a list
+  field), keeping the keys it was given as given, so a `select`/`columns` in them still narrows
+  the parent shape. Its argument is typed by that query when the field's type names a model, so
+  literals such as `select: { title: true }` are kept. It used to return its argument's type.
+- One `PathSegment` type (`string | { name: string; type?: string }`) for `queryFromInfo` paths,
+  `nestedSelection` paths, and `IndirectInclude`, exported from `@pothos/selection-mapper` and
+  re-exported by both plugins. A `{ name, type }` segment given to `nestedSelection` now pins the
+  implementation the field is found under at runtime, as it already did for `queryFromInfo`.
+- The `query` a `t.relation` fallback `resolve` receives (prisma) carries the relation's own
+  arguments (`where`, `orderBy`, `take`, ...) alongside the planned `select`/`include`, so it
+  spreads into a prisma call without a cast. Exported as `QueryFromRelation`.
+- `t.relationCount` (prisma) and `t.relatedCount` (drizzle) accept only list relations.
+- `t.relatedField` (drizzle) accepts the ordinary field options (`deprecationReason`,
+  `extensions`, `authScopes`, ...); its `resolve` may be async and receives the resolve `info`.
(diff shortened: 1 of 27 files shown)


hayes:
Superseded by #1676, the same branch opened directly against `main` (this PR could not be retargeted while GitHub considered it part of a stack). Review happens there.

PR #1676, opened 2026-09-09
feat(prisma, drizzle): shared selection mapper with fragment, variant, loader and async fixes

### LINKED ISSUE #1668: `plugin-prisma`: are `{ name, type }` path segments supported by `queryFromInfo`? Types say `string[]`, runtime normalises objects

hayes:
## Summary

Closes #1668.

`@pothos/plugin-prisma` and `@pothos/plugin-drizzle` each carried their own walker from a GraphQL
selection set to an ORM query. The two were near-mirrors that had drifted, and most of the bugs
found here were in one of them and not the other. This PR replaces both with one package,
`@pothos/selection-mapper`, and fixes every selection-planning bug found on the way.

The branch got here through many rounds — extracting the shared package, making the adapter
contract generic, splitting traversal from accumulation, inverting to collect-then-play, converting
to classes, a vocabulary rename, a bloat audit and a surface shrink. **Only the end state is being
reviewed.** The history is in the commits; the reading order below says which of it is worth
opening. The intent is one release.

## What changes for a user of the two plugins

Every change is in the ten changesets under `.changeset/`, grouped by topic rather than by plugin.
In brief:

**Fragments and paths.** Fields selected behind a fragment that narrows an interface or union, or
under a fragment on an interface the type implements, are planned instead of falling back to a
query per row. Every occurrence of a field under one response key is planned, not only the first.
A relation selected under both `nodes` and `edges { node }` is answered from one query.
`nestedSelection` with a path into a fragment no longer produces an empty selection. `@skip` /
`@include` are honoured on fragments. `queryFromInfo` accepts `{ name, type }` path segments and
returns the caller's own selection when the paths match nothing.

**Variants and type-level selections.** Entering another type of the same model through a fragment
merges that type's type-level selection, order-independently. Two type-level selections that
disagree on a relation's arguments (or, in drizzle, on an `extras` function) are a validation error
naming both types and the entry.

**Loader fallback.** A field loaded through the model loader also carries the parent type's
type-level selection, since the loaded row replaces the parent the resolver sees. Drizzle
`t.relation`, `t.relatedCount` and `t.relatedConnection` load through the loader when the parent row
came from a resolver that fetched it itself, instead of returning `undefined` or an empty page.

**Connections.** Drizzle `relatedConnection` `totalCount` applies the field's `where` (new option
`drizzle.filterConnectionTotalCount`, default `true`). A connection selected through several
fragments plans both the count and the rows. A prisma connection selecting only `pageInfo` loads
only cursor columns. Drizzle `relatedConnection` resolves with the same `pathInfo` it was planned
with, and no longer silently discards a `resolve` option.

**Async selection callbacks.** A field's `select`, a relation `query`, a count `where` and the
connection helpers' callbacks may return promises, and async argument mappers are awaited before a
`select` runs. Prisma previously accepted an async `select` and silently dropped it. A schema with
no async callbacks creates no promise and no closure it did not create before.

**Types and docs.** `queryFromInfo` and `nestedSelection` are typed by what they return. One
`PathSegment` type is shared by `queryFromInfo`, `nestedSelection` and `IndirectInclude`. New
`QueryFromRelation`, `RelatedSelectionFieldOptions`, `PrismaQueriedShape` and `DrizzleQueriedShape`.
`relationCount` / `relatedCount` reject to-one relations in the types. Every option the audit found
undocumented is documented.

**Smaller fixes.** Prisma `onNull` receives `args`, `context` and `info`. `undefined` keys in a
relation `query` compare as absent. `drizzleConnectionHelpers` works without its options argument.
`_count: true` is kept in type-level and field-level selects. `pathInfo.segments[].isList` is true
for non-null lists. `drizzleInterfaceField(s)` resolves a table name to the interface ref. Prisma
caches the per-row fallback plan.

## The package

`packages/selection-mapper`, 2,676 source lines in nine files, ten runtime exports. It is an
implementation detail of the plugins: its README says so, and its versions carry no compatibility
promise of their own.

| file | lines | holds |
|---|---|---|
| `adapter.ts` | 93 | `Adapter<Model, Query, NodeType>`, the ORM boundary. Six abstract members, four inherited merge rules. |
| `node.ts` | 331 | The `Node` tree (model, args, columns, relations, extras) and `NodeAdapter`, the merge/compare/conflict rules over it — the logic each plugin used to write for itself. |
| `plan.ts` | 461 | The `Plan` class: the merges a traversal collected, the fold that plays them into a node, and the two entry points as statics. |
| `walk.ts` | 659 | The traversal. Types, variants, fields, fragments, nested selections, the async chain, the mapping records. |
| `matches.ts` | 678 | Reading a selection set: fragment classification, indirect-include path matching, `@skip`/`@include`/`@defer`, `selectedFieldNames`. |
| `loader-map.ts` | 252 | The two mapping tiers — one per response path, one per row where two plans disagree — and the response-path key cache. |
| `types.ts` | 111 | `Position`, `MergeOptions`, `SelectFn`, `WalkedType`, `EntryOptions`. |
| `deep-equal.ts` | 61 | Structural equality for query arguments, treating `undefined` keys as absent. |
| `index.ts` | 30 | The ten runtime exports and the types beside them. |

**The shape.** A `Plan` is built by `Plan.fromInfo` (the field a resolver was called for) or
`Plan.forParentRow` (the row a field is loaded for). A plan holds no node: the walk collects a list
of merges — a type's selection, a variant's, a nested selection's relation query, a field's — and
decides nothing. `plan.play(seed)` folds that list into a fresh node in order, and that is the only
place a merge is accepted or rejected and the only place a loader mapping is recorded. `plan.query()`
plays and emits. Because the traversal decides nothing, a field that lost a conflict to another
occurrence of itself is still on the list, so a play whose seed makes it fit takes it: `plan.query(select)`
is the same query, and the same mappings, as a plan walked with `select` as its `initial`.

**The contract a new ORM author implements.** Subclass `Adapter<Model, Query, NodeType>` and answer
six members: `modelFor`, `typeSelection` and `fieldSelection` translate the schema into the ORM's own
`Query`; `create`, `merge` and `emit` accumulate. The traversal reads nothing of a node but its
`model`, and nothing of a query at all. Four more members are the merge rules, each inherited with
the answer an adapter that never shares a slot between two consumers wants — `accepts` is true,
`conflict` is none, `absorb` and `acceptsFrom` round-trip through `emit` — so such an adapter writes
six methods and stops. An ORM whose query is a tree of columns, relations and extras extends
`NodeAdapter` instead and writes only `read` (the key loop of its own query, reported entry by entry
to a visitor) and `emit`, plus `extraConflicts` when its extras are not compared by value. That is
what prisma (`src/util/adapter.ts`, 207 lines) and drizzle (`src/utils/adapter.ts`, 155 lines) do;
each plugin's `map-query.ts` is a 130- and 90-line wrapper over the entry points. `@pothos/plugin-prisma-next`
(#1679) is the third adapter, and it is a builder-style ORM with one slot per consumer, so it
extends `Adapter` directly — that is what the generality is for.

## Reading order

Nobody should read 123 commits. Read the end state:

1. **`.changeset/`** — ten entries, the behaviour spec. Start here.
2. **`packages/selection-mapper/README.md`** — the vocabulary and the adapter contract, ~120 lines.
3. **`src/index.ts`**, then **`src/adapter.ts`** — the whole surface and the whole ORM boundary in
   123 lines between them.
4. **`src/plan.ts`** — collect-then-play, the two entry points, `settle`/`reusable`. This is where
   the design lives.
5. **`src/walk.ts`**, then **`src/matches.ts`** — the traversal, then how a selection set is read.
   Long, but they are the two files a bug will be in.
6. **`src/node.ts`** and **`src/loader-map.ts`** — the merge rules and the mapping tiers.
7. **`packages/plugin-prisma/src/util/adapter.ts`** and **`packages/plugin-drizzle/src/utils/adapter.ts`**
   — the contract met twice, side by side.
8. The plugin diffs for the field builders, model loaders and connection helpers.
9. Test names, reading bodies only where a name surprises.

Commits worth opening individually:

| commit | why |
|---|---|
| `566dcac6` | the package's first form — the walker, before any of the reshaping |
| `38ed8f88`, `012faaf3` | the two plugin adoptions; the diffs that delete the mirrored walkers |
| `97e9ce64` | one fragment classification rule (below) — a real behaviour decision |
| `ad53b77c` | the async chain: one optional pending promise, only merges chained onto it |
| `9b0f47b7` | the loader fallback carrying the parent type's selection |
| `a9d5ca3d` | the inversion to collect-then-play |
| `0082f17a` | the adapter and the plan become classes |
| `ccef8eda` | the regression fix (below), with the test that pins it |
| `1c4a9e13` | the per-row fallback plan cache |

Mechanical, skim or skip: the accumulator arc (`61fc27c9`, `cb22a0e1`, `e3965487`, `cc0247ae`,
`709ce8ef`, `9d8af279`) is superseded by `0082f17a`; `2eb806b4` is a rename; `b33247b8` splits a
file that was later merged back; `09d00188` and `baab3200` are a surface shrink and a rename; the
`refactor(selection-mapper)` run at `6cfa68e6`..`1ac6aea7` is dead-parameter removal.

## The evidence

**Behaviour comparison.** The walkers are replaced, so the code diff cannot show behaviour deltas —
the tests can. This branch's plugin test files were run against `main`'s plugin sources (main's
`src` swapped in under this branch's tests, in one worktree, nothing else changed):

| | tests that ran on main | pass | fail | files that could not load |
|---|---|---|---|---|
| prisma | 164 of 167 | 101 | 63 | `count-all` |
| drizzle | 193 of 203 | 134 | 57 | `interface-fields-by-table`, `relation-loading` |

235 tests pass on both — behaviour that did not move; every query those tests issue is the same
before and after. The 120 that fail are the changesets, one by one. Most are new tests for a
behaviour that was wrong or unpinned before; the rest are pre-existing tests whose expectations
were deliberately changed, from five rules (larger fallback queries under E-2, one variant
switching to include mode, `queryFromInfo` returning the caller's selection, one fewer query from
full-path mapping keys, and drizzle's filtered `totalCount`).

To reproduce: from a checkout of this branch, replace `packages/plugin-{prisma,drizzle}/src` with
`origin/main`'s and run `pnpm vitest run` in each.

**Two correctness bugs on main, both silent.** Neither raises an error on main; both are pinned by
tests here.

- *A connection selected in two places returns no edges.* `postsConnection(first: 2) { totalCount }`
  under one fragment and `postsConnection(first: 2) { edges { node { id } } }` under another: main
  plans the connection as totalCount-only from the first occurrence it meets and answers the second
  with **`edges: []`** alongside a correct `totalCount`. Both plugins.
  (`tests/merged-field-nodes.test.ts`.)
- *A variant entered through a fragment loses its type-level select.* A resolver on the variant that
  reads a column its type-level `select` asked for gets **`null`** — `reverseName: null`,
  `email: null` — because the row was loaded without it. Prisma.
  (`tests/variant-fragments.test.ts`.)

**A regression this branch introduced and fixed.** Loader mappings are keyed by response path with
list indices dropped, so one key serves every row of a list. That holds only while every row was
loaded by the same plan. A list can return rows of mixed loaded state; the rows the planned query
did not load fall back to the model loader, which plans the same fields again and, for a while on
this branch, recorded its mappings *over* the plan's at the shared key. A `relatedConnection`
beneath a sibling row then paged with the fallback's arguments against rows the planned query had
fetched — a different edge set and a wrong `hasNextPage` for that one row, with no error and no
extra query. Fixed in `ccef8eda`: where two plans disagree about a key, the second is recorded
against its own row. Pinned by `packages/plugin-drizzle/tests/mixed-loaded-rows.test.ts`, which
asserts that a fallback row and its planned siblings each page at their own position, including
when a sibling completes a tick later. This never reached a released version: main's loader mappings
carry no position at all, so a row there cannot page with a sibling's.

**Benchmark.** Ten (prisma) and eleven (drizzle) documents, planning measured in isolation and
end to end, six reps a side, interleaved in one worktree with main's plugin `src` swapped in
between reps; identical schema hash both sides; the queries issued are identical on every case.
`min` of 400 planning calls, median across reps:

| prisma | main → branch | Δ | | drizzle | main → branch | Δ |
|---|---|---|---|---|---|---|
| flat scalars | 1.21 → 1.75 µs | +45% | | flat scalars | 4.46 → 3.46 µs | −22% |
| nested relations (3 deep) | 10.29 → 13.00 | +26% | | nested relations (3 deep) | 16.71 → 14.75 | −12% |
| relation with args | 5.29 → 7.13 | +35% | | relation with args | 6.00 → 6.46 | +8% |
| same relation twice (fragments) | 14.08 → 14.71 | +4% | | same relation twice (fragments) | 20.46 → 17.13 | −16% |
| connection nodes+edges | 6.46 → 6.92 | +7% | | connection nodes+edges | 8.79 → 8.33 | −5% |
| connection totalCount only | 0.92 → 0.58 | −36% | | connection totalCount only | 1.17 → 0.67 | −43% |
| union variants | 7.33 → 8.71 | +19% | | interface variants | 5.21 → 5.25 | +1% |
| errors wrapper (indirect) | 5.88 → 7.04 | +20% | | errors wrapper (indirect) | 7.33 → 7.13 | −3% |
| 50 rows, per-row resolvers | 8.13 → 8.58 | +6% | | 100 rows, per-row resolvers | 8.96 → 8.75 | −2% |
| model loader fallback | 7.21 → 8.42 | +17% | | conflicting args (per-row fallback) | 9.83 → 12.00 | +22% |
| | | | | node loader (findUnique) | 11.50 → 8.42 | −27% |

**Drizzle plans faster on eight of eleven documents. Prisma plans slower on nine of ten**, by 4–45%,
with allocation up by a similar margin. End-to-end this is invisible — planning is 0.1–3% of a
request that talks to a database, and every e2e delta measured is inside the noise — but it is a
real regression against main and it is not what an earlier run of the same benchmark showed. An
intermediate build of this branch (before `0082f17a`, the class conversion) planned prisma
documents *faster* than main on nine of ten; the same harness on the tip does not. The class
conversion is the prime suspect and has not been bisected. See "look at hardest".

## What to look at hardest

- **The prisma planning regression above.** It is the one measured number that got worse, it
  appeared during a refactor with no behaviour change attached to it, and nothing in the branch
  guards against it. Whether it is worth a bisect before merge, or worth accepting given the e2e
  numbers, is a maintainer call.
- **`plan.ts`: `settle`, `reusable` and `enterParentType`.** A settled play is reused whole by a
  later play whose seed conflicts with nothing in it. The argument for why that is sound (every
  rejection the settled play made was against something the seed leaves in place) is written in
  the doc comment; it is the subtlest thing in the package.
- **One fragment classification rule** (`97e9ce64`). A walk on an object type is a walk on rows of
  that type — the field's own type, a `typeName`, a pinned `nestedSelection` type, or a node load —
  so a fragment on another object type of the same model is not entered; it is entered while
  walking an interface, where rows may resolve to it. Prisma used to enter it whenever the field's
  declared type was abstract, planning a sibling variant's selection for rows that could never be
  it. One prisma expectation changed for this.
- **`loader-map.ts`, the two tiers.** The shared tier answers for every row of a field; the per-row
  tier exists only where two plans disagree about a key. `disagreed` stays false for every request
  whose rows were all loaded the same way, so the row tier costs those nothing. Read `claim()`.
- **Bump level.** Both plugins are `minor`. Two type changes stop code compiling that previously
  failed at runtime (`relationCount`/`relatedCount` on a to-one relation; reading `.select` off a
  `queryFromInfo({ select })` result, now optional), and there is a new drizzle option, new exports
  and async support. `@pothos/selection-mapper` is published for the first time.

## Open items

- **A fallback resolver's returned rows are not stamped.** When a prisma `t.relation` has a
  user-supplied `resolve` and the planned query did not load it, the resolver fetches its own rows;
  nothing calls `setRowMappings` on what it returns, so resolvers beneath those rows read the shared
  tier rather than a mapping of their own. `loader-map.ts` documents the answer for a row nothing
  can be told about: the resolvers beneath fall back and load their own data.
- **Prisma's fallback path writes the shared tier with no row to attach to.** `fallbackQueryFromInfo`
  plays its plan and records through `setLoaderMappings`, which replaces at the key rather than
  claiming per row. This is harmless today because prisma never reads a mapping's `position` —
  only drizzle does, for `pathInfo`. If prisma ever surfaces positions, this becomes the same class
  of bug `ccef8eda` fixed for the model loader.
- **`async` declared types.** `queryFromInfo` keeps a synchronous declared type and returns a
  promise only when a callback beneath the field is async. A resolver that spreads that promise into
  a prisma call gets an empty query and silent fallback loading. The docs say to await it; nothing
  enforces it.
- **Pre-existing, out of scope:** an object inheriting a field from two model interfaces looks its
  loader mapping up under the wrong interface and falls back; a plain `objectType` implementing a
  model interface has no loader and throws a `TypeError` if a relation misses its mapping.

## Verification

At the tip: `@pothos/selection-mapper` 147 tests, `@pothos/plugin-prisma` 167, `@pothos/plugin-drizzle`
203; `pnpm type` and biome clean in all three; `pnpm changeset status` reports a minor for each of
the three. Root `pnpm run ci` passes except `plugin-prisma-utils#test`, which needs its Postgres
container. Each layer of the stack was reviewed by a separate agent run with fresh context.


packages/selection-mapper/src/loader-map.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { GraphQLResolveInfo } from 'graphql';
+
+/**
+ * What a walk records for one field whose selection was merged: the mappings of the fields
+ * walked beneath it, and the adapter-owned `extra` its select function was called with (drizzle's
+ * `PathInfo`), so the resolve path can find the same value the select path used.
+ */
+export interface Mapping {
+  nested: Mappings;
+  extra?: unknown;
```
hayes: should we make this generic?

Issue #942, opened 2023-06-22
[HELP] prisma-relay and Federation: Non-shareable field errors


chrisgate:
I am using @JipSterk  example project to navigate my way in using those technologies

**Project**  [[pothos-federation-prisma-relay](https://github.com/JipSterk/pothos-federation-prisma-relay)](https://github.com/JipSterk/pothos-federation-prisma-relay)
I have below errors whenever I added more than one **prismaConnection** in diff service

**accounts-service**
builder.queryField("users", (t) =>
  t.prismaConnection({
    type: "User",
    cursor: "id",
    resolve: (query) =>
      db.user.findMany({ ...query, orderBy: { createdAt: "desc" } }),
  })
);

**comments-service**
builder.queryField("comments", (t) =>
  t.prismaConnection({
    type: "Comment",
    cursor: "id",
    resolve: (query) =>
      db.comment.findMany({ ...query, orderBy: { createdAt: "desc" } }),
  })
);

**@acme/gateway:codegen:supergraph:     UNKNOWN: Non-shareable field "PageInfo.hasNextPage" is resolved from multiple subgraphs: it is resolved from subgraphs "accounts-service" and "comments-service" and defined as non-shareable in all of them
@acme/gateway:codegen:supergraph:     UNKNOWN: Non-shareable field "PageInfo.hasPreviousPage" is resolved from multiple subgraphs: it is resolved from subgraphs "accounts-service" and "comments-service" and defined as non-shareable in all of them
@acme/gateway:codegen:supergraph:     UNKNOWN: Non-shareable field "PageInfo.startCursor" is resolved from multiple subgraphs: it is resolved from subgraphs "accounts-service" and "comments-service" and defined as non-shareable in all of them
@acme/gateway:codegen:supergraph:     UNKNOWN: Non-shareable field "PageInfo.endCursor" is resolved from multiple subgraphs: it is resolved from subgraphs "accounts-service" and "comments-service" and defined as non-shareable in all of them**

... I have had those  UNKNOWN: Non-shareable fields errors in my real project but I decided to clone @JipSterk project to see what I missed

I am hoping @hayes can help with the missing piece.

Thanks

chrisgate:
[From Apollo Website](https://www.apollographql.com/docs/technotes/TN0029-relay-style-connections/#can-i-use-relay-style-connections-with-apollo-federation)

how can I achieve below with our lovely Pothos
`type PageInfo @shareable {
  hasNextPage: Boolean!
[...]

PR #1678, opened 2026-09-09
feat(plugin-prisma-next): initial plugin for the prisma-next ORM

hayes:
## Summary

Initial version of `@pothos/plugin-prisma-next` — a Pothos plugin for the new prisma-next ORM. This is PR #1627 restacked onto `feat/docs-site-playground` (#1677) and brought up to `@prisma-next/* 0.16`. The description below starts from #1627's and is updated where the plugin has moved on since it was written (the plugin is publishable, the `postinstall` clone-build hook is gone because prisma-next is on npm, and N:M relations are supported — see the changeset).

### What's in

**Schema-builder surface**
- `builder.prismaObject` / `builder.prismaInterface` / `builder.prismaNode`
- `builder.prismaObjectField(s)` / `builder.prismaInterfaceField(s)` for cross-file extension
- `t.prismaField` / `t.prismaFieldWithInput` (top-level prisma-backed fields)
- `t.prismaConnection` (Relay top-level connection over a contract model)
- On `PrismaNextObjectFieldBuilder`:
  - `t.relation` (auto-included from the selection set)
  - `t.relatedConnection` (paginated relation under a parent type)
  - `t.relationCount` / `t.relationAggregate`
  - `t.variant` (re-expose the same row under a different prismaObject)
  - `t.expose*` family
  - `t.withAuth` (scope-auth interop)
- `prismaConnectionHelpers` for custom paginated fields

**Auto-include walker**
- Walks `info` and emits the orm-client chain (`.select / .include / .combine`) inline onto whatever Collection the user returns
- Honors `pothosIndirectInclude` at type and field level — interop with `plugin-errors` result-unions and `t.variant` same-row descents
- Precomputes per-type relation + column metadata in `onTypeConfig` so the per-resolve hot path never touches the contract
- Per-field overlay lifts namespaced combine slots onto flat parent properties for downstream resolvers
- Validates select shapes (typos in declarative keys throw at request time with a useful message)
- N:M relations resolve directly: prisma-next emits the junction join from the contract's `through` block, so no explicit join model is needed (`tests/prisma-next-m-n-upstream-pin.test.ts` pins the upstream behaviour)

**Connections**
- Cursor pagination (single or compound), forward + backward
- `totalCount` as either a synthetic count branch or a user-supplied callback, gated on whether the client actually selects it
- Edge `cursor` is a memoizing non-enumerable getter so clients selecting only `nodes` skip the per-row encode

**Resolver API**
- Resolvers return an orm-client `Collection` (or a row / row[] / null); plugin auto-detects the Collection case and applies the selection mapper + materializes via `.all()`
- Injects `.take(1)` for single-row `t.prismaField` returns to avoid full-table fetches

**Refs and brands**
- Drizzle-shape ref cache (lazy creation; variants get fresh refs)
- No auto-branding — `ref.addBrand` is exposed for the abstract-position cases that need it (matches `plugin-prisma`)
- `prismaNode` brands rows in its batch loader so the brand-check `isTypeOf` fallback works

**Cross-plugin interop**
- `plugin-relay`: `prismaNode` + `t.prismaConnection` / `t.relatedConnection`
- `plugin-errors`: result-union types descend via `pothosIndirectInclude`
- `plugin-scope-auth`: `t.withAuth` on the field-builder
- `plugin-with-input`: `t.prismaFieldWithInput`

### What's deliberately out

- Auto `exposeDescriptions` — the prisma-next contract doesn't carry descriptions.
- Mongo support — orm-client's Mongo Collection has a different shape; plugin is SQL-only.
- Adopting the shared `@pothos/selection-mapper` — the plugin keeps its own walker in `src/utils/apply-selection.ts`; switching it over is separate work.

### Docs

- Website docs under `website/content/docs/plugins/prisma-next/` (index, setup, objects, relations, interfaces, relay, connections, variants). #1677 excluded #1627's docs-content rewrites as unready, but these pages are a new plugin's only documentation, so they are taken here. They follow main's content layout (a directory with `index.mdx` + `meta.json`, like `plugins/prisma`); the plugin is added to the docs plugin index and the marketing plugin catalog (`components/plugins/plugins.ts`); the only intra-doc links are relative (`./connections`, `./variants`) and resolve to pages in the same directory.
- `packages/plugin-prisma-next/ARCHITECTURE.md` — maintainer-focused walkthrough
- `packages/plugin-prisma-next/REVIEW.md` — design-review decisions
- `packages/plugin-prisma-next/feedback/*.md` — upstream-feedback summaries

## How this was restacked

`mh--plugin-prisma-next` (93 commits, 411 files) is the plugin plus a stale copy of the playground/docs-site work, so it could not be rebased. The branch was built from path-filtered checkouts of the source on top of `origin/feat/docs-site-playground` (#1677), in five commits:

1. `feat(plugin-prisma-next)` — `packages/plugin-prisma-next/**` (62 files) and `.changeset/initial-prisma-next.md`, adapted to main's TypeScript 7 build.
2. `docs(plugin-prisma-next)` — the plugin docs and `website/playground-examples/prisma-next-plugin/**`.
3. `feat(website)` — the playground support, ported by hand (below).
4. `chore(plugin-prisma-next)` — the 0.14 → 0.16 upgrade.
5. `fix(website)` — a `bundle-types.ts` fix found during the upgrade.

**Taken**
- The plugin, its tests and fixtures; the changeset; the docs; the playground example.
- `biome.json`: the `packages/plugin-prisma-next/tests/fixtures/**` ignore, plus ignores for the generated `playground-examples/**/contract.{json,d.ts}` and the sql.js driver shim. The `**/packages/**/client/**` generalisation, the `.next`/`node_modules`/playwright ignores and the `*.worker.ts` override were already on #1677.
- Playground support for running prisma-next in the browser (below).

**Dropped**
- `scripts/esm-transformer.ts` fix — main's TS7 upgrade (a32b270e) deleted the transformer and the `esm:extensions` step in every package. The plugin now uses `tsc` (not `tsgo`), no `esm:extensions`, and `.js` extensions on every relative import in `src/` (biome `useImportExtensions`), like every other package.
- `.gitignore` entries (`!website/lib/`, playwright artifacts) — already on #1677.
- `turbo.json` `@pothos/website#generate` task — the website on #1677 has no `generate` script; its `build` script runs `bundle-types` and `build-examples` itself.
- `playwright.config.ts`, `@playwright/test` and `playwright` — the plugin has no `tests/e2e` and no playwright dependency; the config only drove website e2e specs that #1677 dropped.
- `.claude/skills/agent-browser/SKILL.md` — tooling, unrelated to the plugin.
- Every other website difference between the source and `mh--playground` (102 files) — staleness, left alone.

## Playground support ported

Isolated from `git diff origin/mh--playground origin/mh--plugin-prisma-next -- website ':!website/content' ':!website/playground-examples'` and re-applied onto #1677's newer site code. #1677 had already generalised two of the source's hooks (an example-stub module registry and an extension-panel slot), so the port plugs into those instead of re-adding the prisma-next-specific hooks.

| Hunk | Went into |
| --- | --- |
| Bundle `@pothos/plugin-prisma-next` with the workspace `@pothos/core` (source: replaced the plugin list) | `website/lib/playground/plugins-bundle.ts` — added alongside the existing plugins, plus a side-effect import of `prisma-next-bundle` |
| `@prisma-next/*` runtime modules + synthetic `@pothos/playground-capture` exposed to user code (source: `getPrismaNextModules()` called from `execution-engine.ts`) | `website/lib/playground/prisma-next-bundle.ts` — same module map, registered via `registerExampleStubs()`; `execution-engine.ts` untouched (its bare-specifier `external` resolver and stub lookup already cover `@prisma-next/*`) |
| SQL / Prisma-query-AST capture middleware (source: own capture slot + `useQueryRunner` edits) | `website/lib/playground/prisma-next/{capture,capture-middleware,index}.ts` — panels are pushed into `extension-panels-slot`, which `useQueryRunner` on #1677 already resets/reads; `useQueryRunner.ts` untouched |
| sql.js-backed `@prisma-next/driver-sqlite` shim + `/seed` registry (`runtime.mjs`, `runtime.d.mts`, `seed.mjs`, `seed.d.mts`) | `website/lib/playground/prisma-next/driver-sqlite/` (verbatim) |
| Turbopack `resolveAlias` for `@prisma-next/driver-sqlite/{runtime,seed}` and `transpilePackages: ['@pothos/plugin-prisma-next']` | `website/next.config.mjs` (added; the source's version also deleted #1677's redirects — not taken) |
| `paths` for the shim's `.d.mts` | `website/tsconfig.json` |
| Lazy `ensurePrismaNextTypes()` (extra-libs + `paths` into Monaco on first `@prisma-next/` import), disposal on reset | `website/lib/playground/setup-monaco.ts` |
| Keep ATA away from `@prisma-next/*` (`SKIP_PATH`, `ATA_SKIP_BARE`) | `website/lib/playground/setup-monaco-ata.ts` |
| `plugin-prisma-next` in `PLUGIN_PACKAGES`; `readPrismaNextTypes()` (installed `@prisma-next/*` `.d.mts` → `.d.ts` extra-libs, generated `paths` map, shim special-case, CLI/emitter skip list); `prismaNextTypeFiles` / `prismaNextPaths` exports | `website/scripts/bundle-types.ts` (the source also removed the dataloader/complexity/mocks/sub-graph/add-graphql plugins — not taken) |
| Feed the prisma-next type files, `paths` resolution and the `@pothos/playground-capture` ambient to the example type-checker | `website/scripts/test-playground-examples.ts` |
| `@pothos/plugin-prisma-next`, nine `@prisma-next/*` runtime packages, `sql.js` | `website/package.json` (the source's script changes and plugin removals — not taken) |
| Examples picker category map, `types.ts`, `defaults.ts`, `execution-engine.ts`, `useQueryRunner.ts` | No change needed: #1677 already carries `generated`/`hidden` file flags, `.json`/`.sql` loaders and the stub/panel hooks; the source's versions of these files are older |

`website/playground-examples/prisma-next-plugin/metadata.json` prerequisites now use #1677's example ids (`getting-started-first-schema`).

## prisma-next 0.14 → 0.16

Every `@prisma-next/*` dependency in `packages/plugin-prisma-next/package.json` (peer + dev) and `website/package.json` moves from `^0.14.0` to `^0.16.0` (latest, published 2026-07-21). Sources: the GitHub releases / root `CHANGELOG.md` for v0.15.0 and v0.16.0, the repo's upgrade recipes, and a diff of the 0.14.0 vs 0.16.0 published `.d.mts` for every package the plugin touches.

What changed upstream, and what was applied:

- **Unchanged, verified in the type diff:** `sqlite()` from `@prisma-next/sqlite/runtime` (both overloads and options), `SqlMiddleware` hooks, `AfterExecuteResult`, `SqlExecutionPlan`, `RuntimeDriverDescriptor` / driver instance contract, `Contract` from `@prisma-next/contract/types` and `@prisma-next/sql-contract/types`, `resolveDomainModel`, `@prisma-next/adapter-sqlite/codec-types`, the ORM `Collection` method set and public exports. No plugin source changes were needed; the plugin, its tests and the sql.js shim run as-is.
- **Contract emit (0.16, #989):** foreign keys no longer carry `constraint`/`index` booleans; each backing index is a discrete named `indexes[]` entry. The plugin never read those fields. The generated contracts were re-emitted with the 0.16 CLI (`npx @prisma-next/cli@0.16.0 contract emit`): the test fixture `sample-contract.{json,d.ts}` (from the same blog PSL as the example, minus the `@unique` email — the JSON differs from the old fixture only by the FK/index change and hashes) and the example's `contract.{json,d.ts}` in all five steps. The new `contract.d.ts` gains `StorageColumnTypes` / `StorageColumnInputTypes` and the 6-parameter `TypeMaps`.
- **Include decoding through codecs (0.15, #942):** included relation scalars now decode through their codecs like top-level rows. All 218 plugin tests (which assert on included `createdAt` values) pass unchanged; the example resolves included `createdAt` via `new Date(...)`, which accepts both.
- **`include()` relation-name constraint** became `VariantAwareIncludeRelationNames` — the plugin builds include chains from contract metadata and its type tests still pass.
- **Not adopted:** `IncludeRefinementCollection` / `IsToManyRelation` / `IncludeRefinementResult` are still module-internal in 0.16 (declared, not exported) and the many-to-many cardinality is still `'N:M'`, so the plugin's reconstructions and runtime guard stay; the TODO in `src/types.ts` records the 0.16 check. `@prisma-next/sqlite/static` (0.15) is not used.
- The `_prisma_marker` shape the seeds write (`space`, `core_hash`, `profile_hash`, `contract_json`, `canonical_version`, `updated_at`, `app_tag`, `meta`, `invariants`) matches what the 0.16 sqlite adapter reads.
- Version mentions in the docs callouts, the changeset, the seed/PSL comments and the fixture comment now say 0.16.

Found while upgrading: `bundle-types.ts` resolved transitive `@prisma-next/*` packages by taking the first matching `node_modules/.pnpm` entry, so after the bump half of the embedded type trees were still 0.14.0 (pnpm keeps the old line's directories). It now takes the newest version (commit 5).

## Verification

From the repo root, all green:

- `pnpm install`
- `pnpm turbo run build --filter=@pothos/plugin-prisma-next...` — builds (cjs, dts, esm)
- `packages/plugin-prisma-next`: `pnpm vitest run` — 19 files, 218 tests passed, type errors: none; `pnpm type` — clean
- `pnpm biome check --write packages/plugin-prisma-next website` then `pnpm biome check packages/plugin-prisma-next website` — 430 files, no diagnostics
- `pnpm turbo run build --filter=@pothos/website...` — bundle-types (25 plugin definitions, 190 `@prisma-next/*` type files, 118 paths; every embedded package at 0.16.0), build-examples (62 bundles incl. `prisma-next-plugin` × 5 steps), check-playground-refs, `tsc --noEmit` — all pass
- `pnpm --filter @pothos/website test` — 53 example bundles type-check and validate (all five prisma-next steps included), 10 unit tests pass
- Extra: `pnpm --filter @pothos/website next-build` — compiles and prerenders 157 pages (exercises the turbopack alias and `transpilePackages`)

## Open problems

- The playground was not exercised in a real browser here (no e2e). The bundle/type-check path is verified end to end; the in-browser run (sql.js WASM from jsdelivr, the driver shim, the capture panels) is the same code as on the source branch, re-wired through #1677's stub registry and panel slot.
- The example type-checker skips GraphQL query validation for every stepped example (it looks the queries up by the step id, which is not a directory) — pre-existing on #1677, so the prisma-next `.graphql` files are type-checked only as part of the bundle, not validated against the built schema.
- The docs pages are #1627's content; #1677 excluded that PR's docs rewrites as unready, so they may want the same editorial pass as the rest of the site.
- The plugin keeps its own selection walker rather than `@pothos/selection-mapper` (which is on the base); adopting it is separate work.

## Test plan

- [x] `pnpm --filter @pothos/plugin-prisma-next test` — 218 passing, 0 type errors
- [x] `pnpm --filter @pothos/website test` — 53 examples, 10 unit tests
- [x] `pnpm --filter @pothos/website next-build`
- [ ] Open `/playground?example=prisma-next-plugin` in a browser and run a query: SQL and Prisma query AST panels appear
- [ ] Docs render at `/docs/plugins/prisma-next`


hayes:
merged via a separate PR

PR #1692, opened 2026-09-11
fix(core): preserve resolver literal inference

hayes:
Fields now accept individual enum values and discriminated object-union members from synchronous and asynchronous resolvers without return annotations or assertions. Nested mutable backing arrays retain their mutability, and invalid enum members, missing backing properties, and nullability violations remain errors.

Use the existing resolver return-shape generic to preserve literals through a recursive inference type. Keep the original resolver call signature for generic downstream wrappers and constrain the inference signature's return type so extracting `ReturnType` cannot bypass validation. An async overload supplies promise context while retaining plugin-specific return constraints. The existing `t.field({ type, resolve })` calling API and field execution behavior are preserved.

The PR contains the implementation, patch changeset, and regression tests. Standalone experiments have been removed.

Validation:
- Core: 11 test files, 70 tests passed.
- Core type tests pass on TypeScript 5.9.3 and 7.0.2, including positive inference cases and negative validation cases.
- Core declaration generation passes.
- Relay, Errors, and Scope Auth source and test type checks pass against the updated core declarations.
- Formatting, lint, and whitespace checks pass.


huanglinfei091-cmd:
Hi — I noticed the current `Node.js CI / build (24)` job ends on the standard TypeScript
`TS2345` diagnostic in `src/field-builder.ts`. I maintain RunBack, an early alpha that takes a
failed GitHub Actions URL and tries to reproduce the same structured failure locally.

If this would be useful while the PR is active, could you try it on a Linux Docker host?

```bash
curl -fsSL https://raw.githubusercontent.com/huanglinfei091-cmd/runback/main/scripts/install-release.sh | bash
~/.local/bin/runback doctor
~/.local/bin/runback https://github.com/hayes/pothos/actions/runs/34560373095 --job "build (24)"
```

I have not pre-counted this run as successful. `SAME_FAILURE`, `DIFFERENT_FAILURE`,
`INSUFFICIENT_EVIDENCE`, or a blocked result are all useful. If you try it, a reply with the
verdict, `Stage`/`Cause` when shown, rough time, and any install friction would be enough.
Please do not post tokens, credentials, or private logs. No star or promotion is requested.

